PR #59, opened 2020-09-23
Filter for learn catalog: UI changes

maleekha:
Filter UI changes.

The interaction of the UI with the store can be understood via the following diagram:

updated ui changes


client/src/Features/MicrosoftLearn/MicrosoftLearnFilterCore.ts
```diff
@@ -0,0 +1,217 @@
...
+    const parents = filteredProducts.filter(product => getParent(product) === '');
+    const childrenToInclude = filteredProducts.filter(
+      product => !parents.includes(product) && !parents.includes(getParent(product))
+    );
+    return [...parents, ...childrenToInclude];
+  }
+}
+
+export const getRegexs = (searchTerm: string): RegExp[] => {
+  searchTerm = searchTerm.replace(/\W/gi, '');
```
nehalilani, line 156: should it not be this?
```suggestion
  searchTerm = searchTerm.replace(/\w/gi, '');
```

client/src/Features/MicrosoftLearn/MicrosoftLearnList.tsx
```diff
@@ -57,42 +57,50 @@ const MicrosoftLearnListInner = ({ styles }: IStylesOnly<MicrosoftLearnListStyle
...
-                >
-                  {ListRow}
-                </FixedSizeList>
-              );
-            }}
-          </AutoSizer>
+          <>
+            {filteredCatalogContent && filteredCatalogContent.length>0 &&
+              (<div className={classes.contentCount}>
+              <Text variant="mediumPlus" className='contentCountText'> {filteredCatalogContent.length.toLocaleString()} results from Microsoft Learn </Text>
```
nehalilani, line 72: just curious to know why did we need to use `toLocaleString` here?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentProps.ts
```diff
@@ -0,0 +1,35 @@
+import { Product, LearnType, Level, Role } from '../../Models/Learn';
+import { FilterType } from '../../Models/Learn/FilterType.model';
+import { FormEvent } from 'react';
+import { FilterComponentStyles } from './MicrosoftLearnFilterComponentUtils';
+
+export type LearnTypeName = 'Learning Path' | 'Module';
+
+export type LearnTypeFilterOption = {
```
nehalilani, line 8: any specific reason for creating this?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,139 @@
...
+export type FilterComponentStyles = SimpleComponentStyles<
+  'root' | 'title' | 'search' | 'optionsList' | 'subOptionsList' | 'filterItem'
+>;
+
+type FilterComponentTypes = FilterType.products | FilterType.roles | FilterType.levels | FilterType.types;
+
+export const FilterComponent = (props: { type: FilterComponentTypes; name: string }) => {
+  const { filterStore, catalog } = useStore('microsoftLearnStore');
+  return useObserver(() => {
+    const displayOptions =
```
nehalilani, line 25: we can create a separate function `getDisplayOptions` and/or use a switch case instead of nested conditional statements. Thoughts?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,139 @@
...
+      props.type === FilterType.products
+        ? getProductsToDisplay(filterStore.displayFilter[FilterType.products], catalog?.products)
```
nehalilani, line 27: see if this works, to make it consistent with other lines
```suggestion
        ? getProductsToDisplay(filterStore.displayFilter[props.type], catalog?.products)
```

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,139 @@
...
+        : props.type === FilterType.levels
+        ? getLevelsToDisplay(filterStore.displayFilter[props.type], catalog?.levels)
+        : props.type === FilterType.roles
+        ? getRolesToDisplay(filterStore.displayFilter[props.type], catalog?.roles)
+        : getTypesToDisplay(filterStore.displayFilter[props.type]);
+    if (props.type === FilterType.products) {
+      return (
+        <>
```
nehalilani, line 35: do we need this here?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterItem.tsx
```diff
@@ -0,0 +1,82 @@
+import { FilterItemProps } from './MicrosoftLearnFilterComponentProps';
```
nehalilani, line 1: please include license headers in all newly created files

client/src/Features/MicrosoftLearn/MicrosoftLearnList.tsx
maleekha: It formats the number of course contents (of type `Number`) to a `string` with commas at appropriate places.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentProps.ts
maleekha: The existing `LearnType.model` (~ `LearnTypeDto`), unlike the other three (product, role, level), is a simple string, and not an object with the properties {`id`, `name`}. The same (`LearnTypeFilterOption`) was defined here to make the rest of the implementation of `FilterComponents`/`FilterItems` common for all four.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,139 @@
...
+              } else if (!target.checked && value) {
+                filterStore.removeFilter(type, [value]);
+              }
+            }}
+          />
+        </>
+      );
+    } else {
+    return (
+        <>
```
nehalilani, line 69: 1. please fix the indentation
2. do we need this here ?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: Have made the necessary changes!

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: Had to return a block-level component from the function component, therefore added a fragment on top.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: Fixed the indentations. 
Fragment-- same as the previous comment.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,151 @@
...
+  'root' | 'title' | 'search' | 'optionsList' | 'subOptionsList' | 'filterItem'
+>;
+
+type FilterComponentTypes = FilterType.products | FilterType.roles | FilterType.levels | FilterType.types;
+
+export const FilterComponent = (props: { type: FilterComponentTypes; name: string }) => {
+  const { filterStore, catalog } = useStore('microsoftLearnStore');
+  return useObserver(() => {
+    
+    const getDisplayOptions = () => {
```
nehalilani, line 31: Thanks for this! Can we also get this function out of the return statement?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  : filterStore.removeFilter(props.type, [...subItems, value]);
+              }
+            }}
+            subItemClickHandler={event => {
+              let target = event?.target as HTMLInputElement;
+              let type = FilterType.products;
+              let value = target.getAttribute('aria-describedby');
+              if (target.checked && value) {
+                filterStore.addFilter(type, [value]);
+              } else if (!target.checked && value) {
```
nehalilani, line 72: can we use `else` instead of `else if` here and in line 93?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+  ],
+  search: [
+    {
+      marginTop: `calc(${theme.spacing.s1}*2)`
+    }
+  ],
+  optionsList: [
+    {
+      width: '260px',
```
nehalilani, line 124: please update these values in `calc` format wherever possible

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponent.tsx
```diff
@@ -0,0 +1,85 @@
...
+const FilterComponentInner = (props: FilterComponentProps): JSX.Element => {
+  const learnStore = useStore('microsoftLearnStore');
+  const [filterSearchTerm, setFilterSearchTerm] = useState('');
+  const [displayOptions, setDisplayOptions] = useState(props.filterOption);
+
+  useEffect(() => {
+    const filterBySearchTerm = (stringExp: string) => {
+      if (stringExp && stringExp?.trim() !== '') {
+        let regexs: RegExp[] = getRegexs(stringExp);
+        if (props.filterOption) {
```
nehalilani, line 28: I only see an if block inside the if block. Won't clubbing both the if blocks together give the same result?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponent.tsx
```diff
@@ -0,0 +1,85 @@
...
+          let filteresDisplay = getDisplayFromSearch(regexs, props.filterOption);
+          setDisplayOptions(filteresDisplay);
+        }
+      } else {
+        setDisplayOptions(props.filterOption);
+      }
+    };
+    setDisplayOptions(props.filterOption);
```
nehalilani, line 36: setDisplayOptions is being called in the if and else both cases. Why are we calling it again here?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,244 @@
...
+      setWidth(window.innerWidth);
+    }, 500);
+    window.addEventListener('resize', debouncedHandleResize);
+    return () => {
+      window.removeEventListener('resize', debouncedHandleResize);
+    };
+  });
+
+  return useObserver(() => {
+    return width > 768 ? (
```
nehalilani, line 221: any specific reason to use `768` value? I'm seeing this at more than 1 place

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const [mainIsOpen, setMainOpen] = useState(false);
+  const [productIsOpen, setProductOpen] = useState(false);
+  const [roleIsOpen, setRoleOpen] = useState(false);
+  const [typeIsOpen, setTypeOpen] = useState(false);
+  const [levelIsOpen, setLevelOpen] = useState(false);
+  const [panelIsOpen, setPanelOpen] = useState(false);
+
+  const classes = themedClassNames(styles);
+
+  const setOpen = (main: boolean, products: boolean, roles: boolean, types: boolean, levels: boolean) => {
```
nehalilani, line 74: is there some better way to write this function?
maybe getting a string and then using switch case to set that particular type open? thoughts?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: Yeah, I did that, I guess it was after I committed. Will update the remote shortly.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: The `value` for the `target` could be `null`! An alternative could be to wrap the `if-else` block inside another `if` check for the `value`, but this seemed cleaner than that, no?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: As far as I referred, 768px is the commonly used screen size for tabs. Using the same here to re-render as we re-size. Have added a `const` for the same to make this more readable.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponent.tsx
maleekha: Don't seem to recall why I added this line. Removing this as it isn't really impacting anything!

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponent.tsx
maleekha: Agreed.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
maleekha: I have updated the margins and paddings to use `calc` and `theme.spacings`. Only the `widths` remain in a few places, which is fixed.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: Have tried to simplify this by passing in limited parameters, that should help!

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+const FilterOptionComparer = (a: FilterOption, b: FilterOption) => {
+  if (a && b) {
+    return a.name.localeCompare(b.name);
+  } else if (a) {
+    return 1;
+  }
+  return -1;
+};
+
+const createOptoinArrayFromKeys = (keys: FilterOption[]) => {
```
nehalilani, line 22: please see if this helps
```suggestion
const createOptionArrayFromKeys = (keys: FilterOption[]) => {
```

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterComponentUtils.tsx
```diff
@@ -139,11 +140,17 @@ const FilterComponentStyles = ({ theme }: IThemeOnlyProps): FilterComponentStyle
...
+      margin: theme.spacing.s2,
       selectors: {
         '.collapseSubMenuIcon': {
           display: 'inline',
-          height: 'max-content'
+          height: 'max-content',
+          color: theme.palette.neutralDark,
+          selectors: {
+            ' i':{
+              color:'#323130'
```
nehalilani, line 151: can we specify the color using `theme.palette` ?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterCore.ts
```diff
@@ -153,7 +153,7 @@ export function getUpdatedURIFromSelectedFilters(
 }
 
 export const getRegexs = (searchTerm: string): RegExp[] => {
-  searchTerm = searchTerm.replace(/\W/gi, '');
+  searchTerm = searchTerm.replace(/([.?*+^$[\]\\(){}|-])/g, '');
```
nehalilani, line 156: Why did change? How was the previous one behaving differently than this?

maleekha: A bug was reported recently for this. Basically with `searchTerm.replace(/\W/gi, '')`  we are also removing the spaces between the multiple search terms (if any) and combining them to one. So like `azure cloud` --> `azurecloud`, and this was causing unexpected behavior. Have changed this line to replace only the invalid characters, directly now.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
```diff
@@ -1,78 +1,128 @@
...
+    }
+  }, [learnStore, qsParams]);
+
+  // Filters are re-initialized based on the search params present in the location object (on browser back/forward), and the content gets updated accordingly.
+  // [Note]- This does not get called when filterStore.updateHistory is trigerred because we use H.createBrowserHistory to update the URL.
+  //       URL updation using it does not cause the current page to reload.
+
+  // TODO: Refactor the logic to start using useHistory hook for history management with proper UI update
+
+  useEffect(() => {
```
maleekha, line 48: Added this hook-- using just the `if` statement was re-initializing the filters on screen resizing. With the hook, we are executing the filter initialize call only on location change (history back/forward).

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,242 @@
...
+        <FilterComponent type={FilterType.types} name="Types" />
+      </div>
+    );
+  });
+};
+
+const FilterPaneSmall = ({ styles }: IStylesOnly<FilterPaneStyles>): JSX.Element | null => {
+  const { filterStore, filteredCatalogContent, isLoadingCatalog } = useStore('microsoftLearnStore');
+
+  const [mainIsOpen, setMainOpen] = useState(false);
```
adbindal, line 66: This all seems to be a good useCase of `useReducer`. You'd be able to model the complete state in it. Might help cleanup some code in here as well.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const [typeIsOpen, setTypeOpen] = useState(false);
+  const [levelIsOpen, setLevelOpen] = useState(false);
+  const [panelIsOpen, setPanelOpen] = useState(false);
+
+  const classes = themedClassNames(styles);
+
+  type PanelOptions = FilterType.products | FilterType.roles | FilterType.levels | FilterType.types | 'main';
+
+  const setOpen = (options: PanelOptions[]) => {
+    const open = (opt: PanelOptions) => options.includes(opt);
```
adbindal, line 78: How about doing it in the following way?
```suggestion
    const open = options.includes;
```

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,242 @@
...
+    if (mainIsOpen) return 'Search Filters';
+    else if (productIsOpen) return 'Products';
+    else if (roleIsOpen) return 'Roles';
+    else if (typeIsOpen) return 'Types';
+    else if (levelIsOpen) return 'Levels';
+    return '';
+  };
+
+  const kFormatter = (num: number | bigint | any) => {
+    return Math.abs(num) > 999 ? (Math.abs(num) / 1000).toFixed(1) + 'K' : Math.sign(num) * Math.abs(num);
```
adbindal, line 112: Why have we not considered the sign in case of `K`?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,242 @@
...
+          onClick={() => {
+            filterStore.resetFilter();
+          }}
+        />
+      </div>
+    );
+  };
+
+  const getPanelContent = () => {
+    if (mainIsOpen) {
```
adbindal, line 135: What do we mean by `mainIsOpen`? That should help understand the logic here.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
```diff
@@ -0,0 +1,242 @@
...
+            <Icon iconName="ChevronRight" />
+          </ActionButton>
+
+          <ActionButton onClick={() => setOpen([FilterType.types])} className={classes.mainPanelActionButtons}>
+            <Text className="buttonTitle">Types</Text>
+            <Icon iconName="ChevronRight" />
+          </ActionButton>
+        </>
+      );
+    } else if (productIsOpen) {
```
adbindal, line 159: Moving to `useReducer` above should help consolidate these else statements.

client/src/Core/Components/Navbar/NavbarTop.tsx
```diff
@@ -32,15 +32,21 @@ const NavbarTopInner = ({ styles }: IStylesOnly<NavbarTopStyles>): JSX.Element |
   const history = useHistory();
   const assignmentStore = useStore('assignmentStore');
   const location = useLocation();
+  const { filterStore } = useStore('microsoftLearnStore');
   
   // The following map can be extended to include other searchParams as well in future in case need be
   let uriSearchParamsMap = new Map<string, string>();
-  uriSearchParamsMap.set(pagesDisplayNames.MSLEARN, ''); // replace empty string with store state
+  uriSearchParamsMap.set(pagesDisplayNames.MSLEARN, filterStore.learnFilterUriParam);
```
adbindal, line 39: Please see if the below is a bit better in readability?
```suggestion
  const queryParamsMap= { pagesDisplayNames.MSLEARN:  filterStore.learnFilterUriParam };
  
 function pushToHistory(item: PivotItem) {
   if(item.props && item.props.itemKey && item.props.headerText) {
     const { itemKey, headerText } = item.props;
     const queryParam = uriQueryParamsMap[headerText];
     history.push(`${itemKey}?${queryParam}`)
   }
 }
 
 function handleLinkClick(item?: PivotItem, event?:MouseEvent): void {
   event?.preventDefault();
   if(item) pushToHistory(item);
 }
```

adbindal (commented):
While the things look good, I've placed some queries inline.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+import { Product, Role, Level, Catalog } from '../../Models/Learn';
+import { FilterType } from '../../Models/Learn/FilterType.model';
+import { LearnTypeFilterOption, FilterOption } from './MicrosoftLearnFilterComponentProps';
+import _ from 'lodash';
+import { getParentProduct, scoreRegex } from './MicrosoftLearnFilterCore';
+import { Filter } from '../../Models/Learn/Filter.model';
+
+const FilterOptionComparer = (a: FilterOption, b: FilterOption) => {
+  if (a && b) {
+    return a.name.localeCompare(b.name);
```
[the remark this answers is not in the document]
maleekha: Although our `Dtos` explicit the type to be `string`, can add a `?` to be safe. Thoughts?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+  } else if (a) {
+    return 1;
+  }
+  return -1;
+};
+
+const createOptionArrayFromKeys = (keys: FilterOption[]) => {
```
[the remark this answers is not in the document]
maleekha: The return type is `Map<FilterOption, FilterOption[]>`. Essentially, since `Roles`, `Types`, `Levels` are currently non-hierarchical, to parse the available array of the corresponding `FilterOption`s to a common map format, I am using this function. In retrospect, `createOptionsMapFromKeys` would be a more suitable name.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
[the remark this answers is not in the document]
maleekha: I left a comment above regarding this as well- Added this hook back, because using just the `if` statement was re-initializing the filters on screen resizing. With the hook, we are executing the filter initialize call only on `location` change (history back/forward).

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
```diff
@@ -1,78 +1,128 @@
...
+    if (
+      catalog !== null &&
+      location.search !== '?' + filterStore.learnFilterUriParam && location.search !== filterStore.learnFilterUriParam
+    ) {
+      filterStore.initializeFilters(catalog!!, new URLSearchParams(location.search));
```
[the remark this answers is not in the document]
maleekha: I think it is redundant here, we are anyway checking for `null` above. Removing this.

client/src/Features/MicrosoftLearn/MicrosoftLearnSearch.tsx
```diff
@@ -15,16 +15,18 @@ export type MicrosoftLearnSearchStyles = SimpleComponentStyles<'root' | 'label'
...
   const learnStore = useStore('microsoftLearnStore');
+  const { filterStore, isLoadingCatalog } = learnStore;
 
   const classes = themedClassNames(styles);
   return useObserver(() => {
     return (
       <div className={classes.root}>
         <Label className={classes.label}>Search</Label>
         <SearchBox
-          onChange={(_e, newValue) => learnStore.updateSearchTerm(newValue || '')}
```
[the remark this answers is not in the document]
maleekha: I think you mentioned about adding them in the other PR review. I can remove them is they make it seem too crowded.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <div className={classes.root}>
+        {tagMap.map(tag => (
+          <DefaultButton
+            className={classes.tags}
+            iconProps={{ iconName: 'StatusCircleErrorX' }}
+            text={tag.name}
+            onClick={() => {
+              if (tag.type === FilterType.products) {
+                let subItems: string[] = [...catalog?.products.values()]
```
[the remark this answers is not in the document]
maleekha: I thought we could keep the `filterStore::removeFilter` common for all for filter types, without any separate checks for `products`. Although we can consider moving this in future?

client/src/Core/Components/Navbar/NavbarTop.tsx
maleekha: I have incorporated most of the above-suggested changes, but for one thing- the map declaration in `line 39`: I tried it before also, it doesn't seem to work as it is, throws off a compiler error on my end. Could be the difference between es6 and es5?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: I was trying something else here! It is obsolete here now, removing the sign check!

client/src/Features/MicrosoftLearn/MicrosoftLearnSearch.tsx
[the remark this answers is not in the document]
adbindal: Right, I asked them since we were adding `any` as the type to these entities, which was essentially wrong. Leaving them as is leaves a room for the type system to perform an auto-detection and pick up an appropriate type.

client/src/Core/Components/Navbar/NavbarTop.tsx
adbindal: The problem is what I'm trying to create is not a `Map` but a Javascript Object. Trying to coerce it with a map would essentially tend to fail.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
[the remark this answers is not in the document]
adbindal: Can we try making a call to this function with a.name being `null`? If that passes, it means we'd need safety net here. Else we can leave it as is.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
adbindal: Right. Right. This got slipped from my mind. Thanks for pointing it out.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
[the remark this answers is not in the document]
adbindal: There's a `toMap` utility created in our codebase, does it seems like a usecase for that?

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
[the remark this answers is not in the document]
adbindal: Right, Right. It slipped my mind. Thanks for refreshing my memory.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: Cleaned up all this by using `useReducer`! Thanks for the suggestion.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: I was using different states to handle what content should be rendered on the panel -- now using the reducer as per your previous comment.

adbindal: Great. The intent for asking the question was to understand what main stood here for? The rest were quite clear, I was a bit unclear on this one though.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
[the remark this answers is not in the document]
adbindal: The reason I was looking at it that way is this seems to be a logic thing and remove can happen not just via this `chicklet` but via some other action as well. `FilterStore` at this moment is responsible for making sure right things get selected, added and displayed (via the Utils). Only, remove is the one that is now not completely responsible for, hence, that was feeling a bit out-of-place. We can consider taking this up in future for sure, just wondering if there's anything that seems like a friction in refactor (something we should be careful of when doing, which is why we plan to take it separately)?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+    const parentProducts = products.filter(item => getParentProductMapping(item.id) === '').sort(FilterOptionComparer);
+    parentProducts.forEach(parent => {
+      let children = products.filter(product => product?.parentId && product.parentId === parent?.id && product.id!==parent.id);
+      productParentChildMap.set(parent, children.sort(FilterOptionComparer));
+    });
+  }
+  return productParentChildMap;
+};
+
+export const getRolesToDisplay = (roleId: string[] | undefined, roleMap: Map<string, Role> | undefined) => {
```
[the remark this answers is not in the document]
maleekha: Made the suggested changes.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
```diff
@@ -1,78 +1,128 @@
...
+    }  
+  }, [location]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  const classes = themedClassNames(styles);
+  const [width, setWidth] = useState(window.innerWidth);
```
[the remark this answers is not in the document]
maleekha: I agree with this. I have moved the resizing part into a custom hook, and that too only inside the FilterPane.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
```diff
@@ -1,78 +1,128 @@
...
+  // TODO: Refactor the logic to start using useHistory hook for history management with proper UI update
+
+  useEffect(() => {
+    if (
+      catalog !== null &&
+      location.search !== '?' + filterStore.learnFilterUriParam && location.search !== filterStore.learnFilterUriParam
+    ) {
+      filterStore.initializeFilters(catalog!!, new URLSearchParams(location.search));
+    }  
+  }, [location]) // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
maleekha: Removing the hook and using a simple if statement. I added it back because the resize-hook was triggering it un-desirably. Given that has been moved to Filter Pane, we can have the if statement here as discussed before.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
```diff
@@ -1,78 +1,128 @@
...
+      window.removeEventListener('resize', debouncedHandleResize);
+    };
+  });
+
+  return useObserver(() => {
+
+      return (
+        <PageWrapper title={pagesDisplayNames.MSLEARN}>
+        <div className={classes.wrapper}>
+          {width > TAB_SCREEN_SIZE && <MicrosoftLearnFilterPane />}
```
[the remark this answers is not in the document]
maleekha: For 2- I have made the necessary changes by handling this using the stylesheet itself.
With respect to 1- Although I agree this component now seems a bit more crowded, how about we handle this as part of the good first issue?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPane.tsx
maleekha: I see. Previously, I had 6 hooks, each responsible for handling a particular view of the small screen filter panel:

- `panelIsOpen`: whether the panel should be visible at all.
- `mainIsOpen`: if this were true, the panel's content would have the four filter options for the user to choose from.
- `productIsOpen`, `roleIsOpen`, `typeIsOpen`, `levelIsOpen`: each handling the respective filter contents. 

Hope that makes sense.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPaneStyles.ts
```diff
@@ -0,0 +1,112 @@
...
+        '.buttonTitle': {
+          fontWeight: FontWeights.semibold
+        }
+      }
+    }
+  ],
+
+  filterPanelTabView: [
+    {
+      width: '340px',
```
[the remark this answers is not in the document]
maleekha: I agree; have made the necessary changes.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterPaneStyles.ts
```diff
@@ -0,0 +1,112 @@
...
+  | 'filterPanelHeader'
+>;
+
+export const FilterPaneStyles = ({ theme }: IThemeOnlyProps): FilterPaneStyles => ({
+  root: [
+    {
+      display: 'flex',
+      flexDirection: 'column',
+      marginRight: `calc(${theme.spacing.l1} * 1.6)`,
+      width: '300px'
```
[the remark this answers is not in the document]
maleekha: For width < 300px, we would have switched to the small screen mode with the button rendering the filter side panel. I have used `max-width` there as suggested. I hope that helps?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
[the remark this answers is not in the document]
maleekha: I don't think there's anything, in particular, we'll have to take care of while refactoring; I can do it in this iteration itself. I just had one small concern though- since it is a part of the logic, would we have to get it tested again, or a check from my end that nothing breaks up should do?

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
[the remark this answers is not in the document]
maleekha: Yes, in fact, that's exactly what we need here. Made the necessary changes!

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterUtils.ts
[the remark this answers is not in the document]
maleekha: With a.name as null, it throws an exception. In retrospect, do we even need the `if-else` checks here? `FilterOption` is either `Product`, `Role`, `Level`, or `LearnType`. I think we can remove it.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return useObserver(() => {
+    let tagMap: FilterTag[] = getDisplayFilterTags(
+      filterStore.displayFilter,
+      filterStore.selectedFilter,
+      filterStore.productMap,
+      catalog
+    );
+    return (
+      <div className={classes.root}>
+        {tagMap.map(tag => (
```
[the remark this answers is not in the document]
maleekha: Sorry, I didn't get this?

adbindal (approved):
Thanks for incorporating the comments. changes LGTM 👍.

client/src/Features/MicrosoftLearn/MicrosoftLearnPage.tsx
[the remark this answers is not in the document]
adbindal: For Good first issues, it is recommended for them to be some bugs in the product and not refactors. Leaving refactors for later makes them more and more difficult with complexities getting added with every commit. That said, I'm fine with not taking up this refactor right now.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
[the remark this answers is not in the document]
adbindal: I think we've refactored it fine for now. Moving to `FilterStore` could be done separately.

client/src/Features/MicrosoftLearn/MicrosoftLearnFilterTags.tsx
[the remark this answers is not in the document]
adbindal: In simple terms, are we able to jump to tags via **tab** keys and pressing **enter / space** removes the tag.

Play with https://docs.microsoft.com/en-us/learn/browse/?products=azure%2Cgithub%2Cm365 to get an idea. I'd like the same to work for our case (screenshot below).
image

> Docs Reference: https://www.w3schools.com/html/html_accessibility.asp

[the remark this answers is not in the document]
maleekha: Thanks for the clarification. Yes, in our case as well, we are able to simulate a similar behavior via **tab** and **enter**.

adbindal (approved):

adbindal (approved):

adbindal:
Merged changes from #53 as well.

PR #64, opened 2020-09-29
Bugfix/nelilani/delete APIs to return 204 if no content to delete/bug 92222

nehalilani:
### Issue
Calling a `DELETE` APIs returns `http500` when there is no content to delete.

### Solution
Adding a check, if the entity to delete `== null`, i.e. the entity to delete doesn't exist in the Database, then return `http204 No Content`.


adbindal (approved):
LGTM 👍

PR #69, opened 2020-10-02
Refactoring Deploy.ps1 and Publish.ps1 to reuse Write-Log.ps1

adbindal:
## Description
- As a part of creating `Cleanup.ps1`, we refactored logging into a separate utility script of its own. This PR starts re-using that inside `Deploy.ps1` (used via run.bat) and `Publish.ps1` (can be used for only publishing assets to an existing deployment).
- Also, refactored the code inside `Deploy.ps1` to take parameters via command-line as well and moved out exception handling to be more explicit.

## Example with all parameters
``` powershell
Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass
.\Deploy.ps1 -ResourceGroupName {rg-name} -AppName {app-name} -IdentityName {id-name} -UseActiveAzureAccount -SubscriptionName {subscription} -LocationName {region}
```

## Testing
- [x] Ensure deployment succeeds with no arguments -- no regression
- [x] Ensure deployment succeeds for custom Resource Group, Identity Name and App Name
- [x] Ensure deployment succeeds when reusing the existing Azure Account (no explicit sign-in)
- [x] Ensure deployment succeeds by passing subscription via command-line (no input other than region)
- [x] Ensure deployment succeeds by passing region via command-line (no input)

### DIFF
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -1,3 +1,8 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+# --------------------------------------------------------------------------------------------
+
 [CmdletBinding()]
 param (
     [string]$ResourceGroupName = "MSLearnLTI",
--- a/deployment/Deploy.ps1
+++ b/deployment/Deploy.ps1
@@ -3,7 +3,8 @@ param (
     [string]$ResourceGroupName = "MSLearnLTI",
     [string]$AppName = "MS-Learn-Lti-Tool-App",
     [string]$IdentityName = "MSLearnLTI-Identity",
-    [switch]$UseActiveAzureAccount
+    [switch]$UseActiveAzureAccount,
+    [string]$SubscriptionName = $null
 )
 
 process {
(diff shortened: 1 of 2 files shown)


adbindal:
## Code Review Tips
- Look through the changes commit by commit (probably in reverse order), that should isolate the changes to logical entities.
- In case the diff seems quite big, **`Hide whitespace changes`** should be a good option to apply as that might streamline the differences.
image

deployment/Deploy.ps1
```diff
@@ -102,7 +102,7 @@ process {
         $SubscriptionList = Get-LtiSubscriptionList
         Write-Log -Message "List of Subscriptions:-`n$($SubscriptionList | ConvertTo-Json -Compress)"    
 
-        $SubscriptionCount = $SubscriptionList.Count
+        $SubscriptionCount = ($SubscriptionList | Measure-Object).Count
```
nehalilani, line 110: why have we included `Measure-Object` here?

deployment/Deploy.ps1
```diff
@@ -122,7 +122,7 @@ process {
         }
 
         $ActiveSubscription = Set-LtiActiveSubscription -Name $SubscriptionName -List $SubscriptionList
-        $UserEmailAddress = $ActiveSubscription.user.email
+        $UserEmailAddress = $ActiveSubscription.user.name
```
nehalilani, line 130: why this change?

deployment/Deploy.ps1
```diff
@@ -3,297 +3,312 @@
...
-    $roleAssignmentOp = az role assignment create --assignee-object-id $identityObj.principalId --assignee-principal-type ServicePrincipal --role $roleName
-    if(!$roleAssignmentOp) {
-        throw "Encountered an Error while creating Role Assignment"
-    }
+        Write-Log -Message "Fetching List of Locations"
+        $LocationList = ((az account list-locations) | ConvertFrom-Json)
+        Write-Log -Message "List of Locations:-`n$($locationList | ConvertTo-Json -Compress)"
 
-    Write-Host 'Role Assignment Created Successfully'
+        if(!$LocationName) {
```
nehalilani, line 140:
```suggestion
        if($LocationName) {
        Write-Log -Message "Using User provided Location Name: $LocationName"
        }else {
            Write-Host "$(az account list-locations --output table --query "[].{Name:name}" | Out-String)`n"
            $LocationName = Read-Host 'Enter Location From Above List for Resource Provisioning'
            Write-Log -Message "User Entered Location Name: $LocationName"
        }


```

nehalilani (commented):
Thank you for the Code Review Tips. 
Changes look good to me. Just a few suggestion/doubts inline

deployment/Deploy.ps1
adbindal: The way I understand it is that `$SubscriptionList` is an iterable and without `Measure-Object`, it does not have any objects. Hence, we could either have done the `foreach` stuff that was happening before, or this.

deployment/Deploy.ps1
adbindal: Are we talking about the `$UserEmailAddress` or the `Set-LtiActiveSubscription` refactor?
- `$UserEmailAddress` was getting previously set as well but that was happening inside a `foreach` loop when we matched a subscription with a name. I simply separated out the 2 things -- matching subscription and using it to update appropriate variables.
-  The refactor was something which I did with a bit longer term goal of moving every part into its own module (like `Write-Log.ps1`). This should allow us to re-use these across `Deploy`, `Publish` and `Cleanup` scripts.

deployment/Deploy.ps1
adbindal: Thanks for the logging suggestion. Committing it right away.

deployment/Deploy.ps1
nehalilani: the change I was talking about is user.`email` -> user.`name`

nehalilani: I see this is something already present in the deploy script!

nehalilani (approved):

adbindal:
> Merging changes with 🤞 .

Issue #67, opened 2020-09-30
Moodle with LTI 1.1


oyara:
I am trying to integrate Learn LTI with my corporate Moodle. I have deployed correctly all the elements in Azure but I cannot continue because I am using LTI 1.1 in Moodle and I don't know where I can find Customer Secret and Shared Secret after finish the deployment. I have all the information for configure with LTI 1.3 but no with LTI 1.1

![image](https://user-images.githubusercontent.com/1382584/94732225-43902600-0366-11eb-92b4-323e7cac766d.png)

leestott:
@oyara Hi so the Key and Secret are from step1. https://github.com/microsoft/Learn-LTI/blob/main/docs/DEPLOYMENT_GUIDE.md when you deploy the service into your Azure Subscription. 

Can you confirm if this resolves your issue

techulabs:
Thanks @leestott 
I thought that I would need to obtain Consumer key and Shared secret from some place in Azure deployment. 
I'll try with "key" and "secret".

Regards.

oyara:
Thanks @leestott 
I thought that I would need to obtain the Consumer key and Shared secret from some place in Azure deployment.

I'll try with "key" and "secret".

Regards.

PR #72, opened 2020-10-02
Update LTI1.1 Consumer Key/Shared Secret values to LearnLTI/LearnLTI

adbindal:
## Description
- Per LSE and github issue #67, it seems like Consumer Key and Shared Secret values of key and secret and not explicit enough in the documentation for onboarding our tool via LTI1.1.
- Hence, making the changes to make these values a bit more explicit, i.e. using **LearnLTI** for both Consumer Key and Shared Secret.

## Testing
- [x] Ensure that we are able to create a new assignment using LTI1.1 Tool and publish it for students.
- [x] Ensure that `GetAllUsers` API is working successfully with new Key/Secret value.
- [x] Ensure that `GetUserDetails` is successful for a valid user.

### DIFF
--- a/deployment/run.bat
+++ b/deployment/run.bat
@@ -1,8 +1,8 @@
+@echo off
 REM --------------------------------------------------------------------------------------------
 REM Copyright (c) Microsoft Corporation. All rights reserved.
 REM Licensed under the MIT license.
 REM --------------------------------------------------------------------------------------------
 
-@echo off
 REM Triggers Learn-LTI Deployment and Publishing Script bypassing the Signing requirements
 PowerShell.exe -ExecutionPolicy Bypass -File .\Deploy.ps1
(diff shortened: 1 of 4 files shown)


deployment/run.bat
```diff
@@ -1,8 +1,8 @@
+@echo off
 REM --------------------------------------------------------------------------------------------
 REM Copyright (c) Microsoft Corporation. All rights reserved.
 REM Licensed under the MIT license.
 REM --------------------------------------------------------------------------------------------
 
-@echo off
```
nehalilani, line 6: Any specific reason for removing this?

adbindal: It's not removed. Just moved to the top of the file. Without this all the `REM` comments on lines 2-4 were ending up getting displayed on the console.

nehalilani: My bad!

nehalilani (approved):

PR #75, opened 2020-10-06
Refactoring azuredeploy

nehalilani:
Refactored the AzureDeploy.json file to reduce redundant code


deployment/azuredeploy.json
```diff
@@ -58,7 +54,16 @@
         "cliArg": "[concat(variables('keyVaultName'), ' ', variables('staticWebsite'), ' ',variables('location'), ' ',variables('ProjectNameSuffix'), ' ',variables('keyName'),' ',variables('resourceGroupName'))]",
         "contributorRoleAssignmentId": "[guid(concat(resourceGroup().id, 'contributor'))]",
         "contributorRoleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'b24988ac-6180-42a0-ab88-20f7382dd24c')]",
-        "readerRoleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'acdd72a7-3385-48ef-bd42-f606fba81ae7')]"
+        "readerRoleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'acdd72a7-3385-48ef-bd42-f606fba81ae7')]",
+        "keyPermissions": [
```
adbindal, line 58: How about using the name `keyVaultPermissions` instead?

adbindal (approved):
LGTM 👍, quite safe to do as well.

deployment/azuredeploy.json
nehalilani: I chose the name KeyPermission matching it with what is shown in the Azure portal. Within a KeyVault, there are 3 types 0f permissions- key, secret and certificate. And we are only giving the "KeyPermissions" to these function apps and not the secret & certificate permissions

adbindal: How about using `KVKeyPermissions` in that case?

nehalilani: updated!

PR #79, opened 2020-10-13
Handling failed post delete service calls

maleekha:
Summary of the changes made:

- Adding a synced-state in each store which gets updated only when post/delete calls succeed, and is also shown to the users on the preview page/student view page.
- In case any post/delete call fails, a sticky message is shown at the top of the page with a pertinent error message based on the service error type. An error is also shown in case the user state is not synced with (even when all the calls succeeded).
- A spinner on the top of the page to indicate that some service calls are in progress.
- Warning inside modal when the user tries to publish the assignment and (I) some service calls are in progress, or (II) the assignment is not synced (due to some error or otherwise).
 - Failure message in case publishStatusChange call fails.


leestott:
Just waiting/scheduling time with Engineering for regression testing to ensure this does not break any aspects.

No Impact to exiting deployment, so users can deploy the services without this change,

PR #83, opened 2020-10-21
Bump react-scripts from 3.4.3 to 3.4.4 in /client

adbindal:
**Reference:** https://github.com/facebook/create-react-app/releases/tag/v3.4.4


leestott (approved):
@adbindal  excellent thanks for applying these changes so quickly

Issue #86, opened 2020-10-27
Not able deploy using the run.bat


manojkharol:
I ran the run.bat script. I was able to successfully login my azure account and later it asked me for my subscription name. I provided the same but there was an error of Invalid subscription name (screenshot attached). After this, it asked for the region and I mentioned eastus2. Later there is an error saying Insufficient Privilege (screenshot attached). 
![image](https://user-images.githubusercontent.com/46344301/97327266-d86d3d00-189a-11eb-80fd-31e221db23b5.png)
![image](https://user-images.githubusercontent.com/46344301/97327460-fc308300-189a-11eb-80b5-9f89701e985c.png)

manojkharol:
The issue is resolved by commenting the validation code from line 96 to 99 in Deploy.ps1 file and adding owner in the Azure Portal.
![image](https://user-images.githubusercontent.com/46344301/97555049-a1fc0300-19fd-11eb-9ad5-a3e2ba49f05c.png). Thanks for the help Aditya! 

However, there is a new issue while Creating Resources in Azure.
![image](https://user-images.githubusercontent.com/46344301/97555292-f4d5ba80-19fd-11eb-9295-3367ffb7797c.png)
Deployment failed. Correlation ID: ab846c6a-0ab2-4fef-9b8d-01a80a577074. {
  "status": "failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'failed'.",
    "details": [
      {
        "code": "DeploymentScriptError",
        "message": "The provided script failed with multiple errors. First error:\r\nERROR: Please run 'az login' to setup account.. Please refer to https://aka.ms/DeploymentScriptsTroubleshoot for more deployment script information.",
        "details": [
          {
            "code": "DeploymentScriptError",
            "message": "ERROR: Please run 'az login' to setup account."
          },
          {
            "code": "DeploymentScriptError",
            "message": "ERROR: Please run 'az login' to setup account."
          }
        ]
      }
    ]
  }
}

adbindal:
@manojkharol, #87 should take care of the `az login` issue that you are currently facing. We've also updated the deployment scripts to start accepting both Subscription ID and Name as user input (see #88). Hope, these changes, once merged, should unblock the deployment on your end.

leestott:
@manojkharol can you confirm this change has fixed your issue.

adbindal:
[...]

PR #87, opened 2020-10-29
Bugfix/Depolyment failure

nehalilani:
## Issue
The deployment was failing while running the Azure Cli scripts as a part of ARM template. The script essentially creates the key in **KeyVault** and enables `$web` Static Website container inside Client Storage Resource.

## Solution
Moved Azure Cli commands for creating the key and enabling static website outside of ARM template, into `Deploy.ps1` file. We've ensured that the ARM template deployment does not suffer due to these missing command executions.
* **KeyVault** key is being used to update `appSettings` for Connect, Users and Platform Function Apps which is not a part of ARM Template.
* Static Website's address is being used to set the **CORS** and `RedirectUrl` in `appSettings`. However, the Url setting does not require Static Website container to be present. It is only required when we are actually publishing client code to Azure which happens inside `Publish.ps1` (after the ARM template).

## Testing

> Dev Sanity Complete

- [x] Ensure that Admin is able to register a new Tool inside LMS using the `/platform` client endpoint.
- [x] Ensure that Teacher is able to create new resource, add necessary contents to it -- all Links, MS Learn, etc., preview it and publish/unpublish it as needed.
- [x] Ensure that Student is able to view the published assignment, access all its contents.


deployment/Deploy.ps1
```diff
@@ -236,13 +236,17 @@ process {
             }
         }
     
-        #Updating the Config Entry EdnaLiteDevKey in the Function Config
-        $KeyVaultLink=$(az keyvault key show --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --query 'key.kid' -o json);
+        #Creating EdnaLiteDevKey in keyVault and Updating the Config Entry EdnaLiteDevKey in the Function Config
+        $ke = az keyvault key create --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --protection software
```
adbindal, line 240: If this fails, should we even proceed with the script 🤔 ?

deployment/Deploy.ps1
```diff
@@ -236,13 +236,17 @@ process {
...
+        $KeyVaultLink = $(az keyvault key show --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --query 'key.kid' -o json);
```
adbindal, line 241: Can we see if this information can be obtained from previous entry itself.

deployment/Deploy.ps1
```diff
@@ -236,13 +236,17 @@ process {
...
         $EdnaKeyString = @{ "EdnaKeyString"="$KeyVaultLink" }
         $ConnectUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.ConnectFunctionName.value $EdnaKeyString
         $PlatformsUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.PlatformsFunctionName.value $EdnaKeyString
         $UsersUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.UsersFunctionName.value $EdnaKeyString
-    
+
+        #Setting index-document and 404-document in StaticWebsite
```
adbindal, line 247: Should we move it to a separate step altogether? What should we do in case this step fails?

deployment/azuredeploy.json
```diff
@@ -470,7 +469,7 @@
                 "AzCliVersion": "2.0.80",
                 "timeout": "PT30M",
                 "arguments": "[variables('cliArg')]",
-                "scriptContent": "az keyvault key create --vault-name $1 --name $5 --protection software;az storage blob service-properties update --account-name $2 --static-website --404-document index.html --index-document index.html;",
+                "scriptContent": "az account list",
```
adbindal, line 472: I would like to remove this step altogether. It is doing nothing at the moment except confusing the reader about its existence. I can understand the urgency of getting this fix in since the current state is a customer impact, however, I'd request you to please follow-up with an immediate PR that has the cleaned up implementation of this change.

adbindal (approved):
The current implementation looks like hot-patching the issue. Given that issue has a customer impact and nobody can deploy Learn-LTI at the moment. I am fine with merging these changes as is to unblock our customers after performing a validation on our end for any regressions. However, I'd request for an immediate follow-up on this fix, performing necessary changes to the current implementation, hardening the edge cases and raising a PR for the same.

> Approving since the changes seem to unblock the failing deployments, without looking to break anything new.

nehalilani:
> The current implementation looks like hot-patching the issue. Given that issue has a customer impact and nobody can deploy Learn-LTI at the moment. I am fine with merging these changes as is to unblock our customers after performing a validation on our end for any regressions. However, I'd request for an immediate follow-up on this fix, performing necessary changes to the current implementation, hardening the edge cases and raising a PR for the same.
> 
> > Approving since the changes seem to unblock the failing deployments, without looking to break anything new.

I agree with you. I'll create a follow-up to incorporate necessary changes as mentioned in your comments

adbindal:
This should fix the deployment issues reported in #86 due to missing `az login` to unblock our customers' deployment.

deployment/Deploy.ps1
nehalilani: updating this in the followup

deployment/Deploy.ps1
nehalilani: it seems like the azure has different commands to create and retrieve key from keyVault

deployment/Deploy.ps1
nehalilani: updating the failure case solve in follow up

PR #89, opened 2020-10-31
Followup: Cleaner implementation for deployment failure fix

nehalilani:
## Description
As a part of moving script execution out of ARM template, we realize that we can perform a much broader set of changes given the new way to deploy. The changes include: -

1. Handle and Report Errors in case of `EdnaLiteDevKey` creation or `$web` Container creation command failures.
2. Remove User Managed Identity and contributor role steps, as those are only required to execute **Az** Script.

>The deployment time has now been reduced to around 1 min 30 sec.

- [x] Ensure that Admin is able to register a new Tool inside LMS using the /platform client endpoint.
- [x]  Ensure that Teacher is able to create new resource, add necessary contents to it -- all Links, MS Learn, etc., preview it and publish/unpublish it as needed.
- [x]  Ensure that Student is able to view the published assignment, access all its contents.
- [x] Re-deploy an RG previously deployed using the main line.


deployment/azuredeploy.json
```diff
@@ -446,54 +413,6 @@
...
-                "AzCliVersion": "2.0.80",
-                "timeout": "PT30M",
-                "arguments": "[variables('cliArg')]",
-                "scriptContent": "az account list",
-                "cleanupPreference": "OnSuccess",
-                "retentionInterval": "P1D"
-            }
-        },
-        {
-            "type": "Microsoft.ManagedIdentity/userAssignedIdentities",
```
adbindal, line 478: No dependency of any stage on `userAssignedIdentities` 🤔?

deployment/azuredeploy.json
```diff
@@ -381,18 +360,6 @@
                 "enablePurgeProtection": true,
                 "tenantId": "[variables('tenantId')]",
                 "accessPolicies": [
-                    {
```
adbindal, line 384: Nice Catch. Thanks.

deployment/Deploy.ps1
```diff
@@ -192,42 +191,16 @@ process {
...
-        Write-Title 'STEP #6 - Creating Managed Identity'
-
-        Write-Log -Message "Creating Managed Identity inside ResourceGroup [ $ResourceGroupName ] with Name [ $IdentityName ]"
-        $identityObj = (az identity create -g $ResourceGroupName -n $IdentityName) | ConvertFrom-Json
-        if(!$identityObj) {
-            throw "Error while creating Managed Identity inside ResourceGroup [ $ResourceGroupName ] with Name [ $IdentityName ]" 
-        }
-    
-        #It takes a few seconds for the Managed Identity to spin up and be available for further processing
-        Write-Log -Message "Sleeping for 30 seconds"
```
adbindal, line 205: Saves some time for the deployment. Bravo!!!

deployment/Deploy.ps1
```diff
@@ -241,19 +214,24 @@ process {
         }
     
         #Creating EdnaLiteDevKey in keyVault and Updating the Config Entry EdnaLiteDevKey in the Function Config
-        $ke = az keyvault key create --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --protection software
+        $keyCreationOp = az keyvault key create --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --protection software
+        if(!$keyCreationOp) {
+            throw "Encountered an Error while creating Key in keyVault"
+        }
         $KeyVaultLink = $(az keyvault key show --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --query 'key.kid' -o json);
```
adbindal, line 221: Can we get rid of this command, and use the output of above command itself?

deployment/Deploy.ps1
```diff
@@ -241,19 +214,24 @@ process {
...
         $ConnectUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.ConnectFunctionName.value $EdnaKeyString
         $PlatformsUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.PlatformsFunctionName.value $EdnaKeyString
         $UsersUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.UsersFunctionName.value $EdnaKeyString
 
-        #Setting index-document and 404-document in StaticWebsite
-        $se = az storage blob service-properties update --account-name $deploymentOutput.properties.outputs.StaticWebSiteName.value --static-website --404-document index.html --index-document index.html --only-show-errors
-
+        #Creating a Container in Static Website
+        $containerCreationOp = az storage blob service-properties update --account-name $deploymentOutput.properties.outputs.StaticWebSiteName.value --static-website --404-document index.html --index-document index.html --only-show-errors
+        if(!$containerCreationOp) {
```
adbindal, line 229: Since this something that can be done manually as well, would it better to also direct the user on how to do this manually 🤔?

deployment/Deploy.ps1
adbindal: Also, since this something that can be done manually as well, would it better to also direct the user on how to do this manually 🤔?

deployment/Deploy.ps1
```diff
@@ -241,19 +214,24 @@ process {
...
         $KeyVaultLink = $(az keyvault key show --vault-name $deploymentOutput.properties.outputs.KeyVaultName.value --name EdnaLiteDevKey --query 'key.kid' -o json);
         $EdnaKeyString = @{ "EdnaKeyString"="$KeyVaultLink" }
         $ConnectUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.ConnectFunctionName.value $EdnaKeyString
         $PlatformsUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.PlatformsFunctionName.value $EdnaKeyString
         $UsersUpdateOp = Update-LtiFunctionAppSettings $ResourceGroupName $deploymentOutput.properties.outputs.UsersFunctionName.value $EdnaKeyString
 
-        #Setting index-document and 404-document in StaticWebsite
-        $se = az storage blob service-properties update --account-name $deploymentOutput.properties.outputs.StaticWebSiteName.value --static-website --404-document index.html --index-document index.html --only-show-errors
-
+        #Creating a Container in Static Website
```
adbindal, line 227: Does it make sense to have separate steps for updating these? 
> Might be a bit better from progress tracking perspective also
1. Step #7 - Updating KeyVault with LTI 1.3 Key
2. Step #8 - Enabling Static Website Container

adbindal (approved):
Changes LGTM 👍. Thanks for the cleanup and making the deployment performant. Let's try to measure the total time of deployment now and add it to the PR.

deployment/Deploy.ps1
nehalilani: Yes that can be done but if the container is not created before the step where client code is installed, the deployment will fail on that step.

deployment/Deploy.ps1
nehalilani: this can be done manually but then the user will also have to set the key in the required functions apps' configs

nehalilani: >Can we get rid of this command, and use the output of above command itself?

I'll see if this can be done somehow

deployment/azuredeploy.json
nehalilani: No, it was only created to give the Managed Identity access to the key vault

vipul-bhojwani:
@nehalilani - This seems like a great improvement over the previous version of the script. 
We may want to add One Test Case each time we release a new version of the deployment script to ensure that re-deploying over an existing deployment (done using the previous version) does not break anything.

adbindal (approved):

PR #90, opened 2020-11-01
X509 Certificate Validation for External Entities

maleekha:
**Changes Made** 

Added server x509-certificate validation for the external entities that our backend connects to, including:

- [Lti Advantage] NRPS Client
- [Lti Advantage] Lti-Launch-Claims Request
- [Lti Advantage] Acces Token Service
- [Lti Advantage] Key Vault Jwt Provider
- [Lti Advantage] Key Vault Pem Provider
- [Lti1] Membership Client
- [Assignment-Learn-Content] Get Learn Catalog

**Test Cases**
- [x] No regression from the user's perspective
- [x] Test by redirecting the calls to a local server, where a broken certificate is installed
- [x] Sanity check on all the LMS


backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Net.Http;
+using System.Security.Cryptography.X509Certificates;
+using System.Net.Security;
+
+namespace Edna.Bindings.LtiAdvantage.Utils
```
maleekha, line 5: @adbindal I had to create a duplicate file for LTIAdvantage. Adding a dependency from `LTIAdvantageBinding` on the other `Utils` project was causing a package version conflict issue (for `System.IdentityModel.TokensJwt`). The version is 6.5.1 and 6.6.0 in the two, respectively.

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Net.Http;
```
adbindal, line 1: Missing Copyright.

backend/Bindings/Edna.Bindings.LtiAdvantage/LtiAdvantageBindingExtensions.cs
```diff
@@ -16,7 +17,7 @@ public static IWebJobsBuilder AddLtiAdvantageBindings(this IWebJobsBuilder build
         {
             builder.AddExtension<LtiAdvantageExtensionConfigProvider>();
 
-            builder.Services.AddHttpClient();
+            builder.Services.AddHttpClient(nameof(EdnaExternalHttpHandler)).ConfigurePrimaryHttpMessageHandler(() => new EdnaExternalHttpHandler());
```
adbindal, line 20: We've added a new line in file **Lti1BindingExtensions.cs** for doing this? Any specific reason for not being consistent across these files?

backend/Bindings/Edna.Bindings.LtiAdvantage/LtiAdvantageBindingExtensions.cs
```diff
@@ -25,5 +26,5 @@ public static IWebJobsBuilder AddLtiAdvantageBindings(this IWebJobsBuilder build
 
             return builder;
         }
-    }
+}
```
adbindal, line 29: Seems like a residual change.

backend/Bindings/Edna.Bindings.LtiAdvantage/LtiAdvantageBindingExtensions.cs
```diff
@@ -5,6 +5,7 @@
 
 using Edna.Bindings.LtiAdvantage.BindingConfigurations;
 using Edna.Bindings.LtiAdvantage.Services;
+using Edna.Bindings.LtiAdvantage.Utils;
```
adbindal, line 8: What is this being used for?

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/AccessTokenService.cs
```diff
@@ -9,6 +9,7 @@
 using System.Net.Http;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using Edna.Bindings.LtiAdvantage.Utils;
```
adbindal, line 12: Same question as above.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/KeyVaultJwkProvider.cs
```diff
@@ -18,8 +20,8 @@ public async Task<JsonWebKey> GetJwk(string keyVaultIdentifier)
         {
             AzureServiceTokenProvider azureServiceTokenProvider = new AzureServiceTokenProvider();
             KeyVaultSecurityKey.AuthenticationCallback keyVaultAuthCallback = new KeyVaultSecurityKey.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback);
-
-            KeyVaultClient client = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(keyVaultAuthCallback));
+            EdnaExternalHttpHandler ednaExternalHttpHandler = new EdnaExternalHttpHandler();
```
adbindal, line 23: What's this change helping out with?

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Net.Http;
```
adbindal, line 1: Any specific reason for keeping this entry tied to `LtiAdvantage` - even though Lti1 is also using it? Maybe we can move it to general utils instead 🤔.

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/LtiAdvantageExtensions.cs
```diff
@@ -23,7 +23,7 @@ public static async Task<ClaimsPrincipal> GetValidatedLtiLaunchClaims(this HttpR
             if (!request.Form.TryGetValue("id_token", out var idTokenValue))
                 throw new NullReferenceException("No ID token is presented in the http request.");
 
-            using HttpClient client = new HttpClient();
+            using HttpClient client = new HttpClient(new EdnaExternalHttpHandler());
```
adbindal, line 26: Why are we not using `_httpClientFactory` here? While all the other things seem to maange `new` internally, having this go against the pattern seems a bit odd.

backend/Functions/Edna.AssignmentLearnContent/Edna.AssignmentLearnContent/Startup.cs
```diff
@@ -23,7 +24,8 @@ public void Configure(IWebJobsBuilder builder)
 
             builder.AddUserBinding();
 
-            builder.Services.AddHttpClient();
+            builder.Services.AddHttpClient(EdnaExternalHttpHandler.Name).ConfigurePrimaryHttpMessageHandler(() => new EdnaExternalHttpHandler());
```
adbindal, line 27: Is it possible to have an `EdnaHttpClient` which does this automatically? That way, we could simply put the following in all the files allowing us to manage the configurations and code hardening at a single place.
``` C#
builder.Services.AddEdnaHttpClient();

// Implementation could be something like below
public AddEdnaHttpClient(this Services services) {
  services
    .AddHttpClient(EdnaExternalHttpHandler.Name)
    .ConfigurePrimaryHttpMessageHandler(() => new EdnaExternalHttpHandler());
}
```

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Net.Security;
+
+namespace Edna.Utils.Http
+{
+    public class EdnaExternalHttpHandler: HttpClientHandler
+    {
+        public static string Name = nameof(EdnaExternalHttpHandler);
```
adbindal, line 9: Shouldn't this be `const` instead?

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+        public EdnaExternalHttpHandler() : base()
```
adbindal, line 10: We may also consider a single static function `CreateDefault` to consolidate all the scattered `new` across the codebase. What this helps is w.r.t passing any additional params or config at the time of creating or performing additional settings if needed inside the `CreateDefault` function itself.

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            ServerCertificateCustomValidationCallback = PerformX509Valiation;
+        }
+
+        public bool PerformX509Valiation ( HttpRequestMessage req, X509Certificate2 cert, X509Chain chain, SslPolicyErrors err )
+        {
+            return (err == SslPolicyErrors.None) && chain.Build(cert);
```
adbindal, line 17: I might need some explanation or pointers on what this is trying to achieve and how it is doing that.

adbindal (changes requested):
Thanks for taking this important set of changes. While the overall approach seems good, I've added some comments inline. Please let me know in case you need more help to align with my POV.

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
adbindal: Hmm. Sounds interesting. Should we consider upgrading those instead 🤔 ?

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/LtiAdvantageExtensions.cs
maleekha: There is some inconsistency in the existing implementation regarding the use of `_httpClientFactory`. At some places, we are creating the `HttpClient` directly, instead of going by the other approach. To stick to the existing one, I added the Handler by passing it as a parameter for the `HttpClient` constructor. Should we this as a follow-up: making use of the factory everywhere?

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
maleekha: I have one implementation in the general utils project as well, which is being used by the Lti1 binding, and the learn-content-api. Adding a dependency for Utils to LtiAdvantageBinding was causing a version conflict. Regarding that, I believe that we should take that (+ refactoring and removing this redundancy) as a separate work item, so that if things break apart, we are able to identify the cause- if it was the version upgrade, or the certificate validation changes. Any thoughts on this?

backend/Bindings/Edna.Bindings.LtiAdvantage/LtiAdvantageBindingExtensions.cs
maleekha: I will make this consistent.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/AccessTokenService.cs
maleekha: Essentially, in order to use the custom handler implementing the server certificate validation logic, we need to pass it a string indicating the `Name` of the handler. All these imports have been made in order to use the `Name`  property of the new Handler (or initialize one in cases where we are using HttpClient directly).

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/KeyVaultJwkProvider.cs
maleekha: Essentially passing the custom `HttpClient` handler as a parameter to the `KeyVaultClient`, so that it uses this instead of the default `HttpClient`.

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
maleekha: I agree. I was trying to make the implementation as consistent as possible with the one done for `EdnaHttpHandler`. But making it const makes more sense, yes. I will update this change.

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
maleekha: I wasn't sure if this was done on purpose, or because of the gap between development periods. The two versions were released pretty close in time though, so that could be a possibility. Should we take that as a separate task, so that we can keep track in case it breaks things up?

backend/Functions/Edna.AssignmentLearnContent/Edna.AssignmentLearnContent/Startup.cs
maleekha: That would be useful! Let me try and get that working, I will push the changes shortly!

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandler.cs
adbindal: Sure. Let's plan a follow-up. Having it in LtiAdvantage is as confusing as it sounds.

adbindal (changes requested):
The build seems to be failing. Please take a look.

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandlerExtension.cs
```diff
@@ -0,0 +1,17 @@
+﻿// --------------------------------------------------------------------------------------------
```
adbindal, line 1: I'm seeing a similar file in Utils folder as well. Any reason for this duplication 🤔 ?

adbindal (approved):
LGTM 👍

backend/Bindings/Edna.Bindings.LtiAdvantage/Utils/EdnaExternalHttpHandlerExtension.cs
maleekha: > The build seems to be failing. Please take a look.

Yes, I have taken care of the same. It's working fine now.

maleekha: > I'm seeing a similar file in Utils folder as well. Any reason for this duplication 🤔 ?

Ideally, they all must only be in the Utils folder only. However, due to a package version conflict (mentioned above), I had to add a duplicate version of the changes. I will right away work on a follow-up task wherein we upgrade the package version in LTIAdvantage, and remove the redundancy.

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
maleekha: Thanks for this comment, I have added modifications to address this! It has made the code a lot cleaner indeed.

backend/Utils/Edna.Utils.Http/Edna.Utils.Http/EdnaExternalHttpHandler.cs
maleekha: The following links should help in explaining the above invocation:
- `X509Chain` `build` method: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chain.build?view=netcore-3.1 
- `X509VerificationFlags` used by the above `build` method to verify the certificates in the chain: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509verificationflags?view=netcore-3.1 (by default: NoFlag i.e. all the checks are performed)

PR #91, opened 2020-11-02
Assinment Links Input Validation

maleekha:
Changes Made:

- Validation for the link entered in the URL field: Must be a valid non-empty URL
- Added character limits for Link Display Text and Description
- The `Add Link`/`Save Changes` button stays disabled as long as any of the above conditions are violated.

Testing:

- [x] Adding invalid data to the client-side input fields: check for validation messages.
- [x] Server-side validation checks using Postman.
- [x] Sanity checks for adding/updating and deleting links.


client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -18,6 +18,24 @@ interface AssignmentLinkEditorProps {
 }
 export type AssignmentLinkEditorStyles = SimpleComponentStyles<'root'>;
 
+const LinkTextValidation = (value: string | undefined, MAXLEN: number) => {
+  if(value && value?.length>MAXLEN){
```
adbindal, line 22: The following might be enough?
```suggestion
  if(value?.length > MAXLEN){
```

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -34,12 +52,15 @@ const AssignmentLinkEditorInner = ({
         value={link?.displayText || ''}
         styles={themedClassNames(textFieldBaseStyle)}
         onChange={(_e, newValue) => onChangeTextField('displayText', newValue || '')}
+        onGetErrorMessage={(value) => LinkTextValidation(value, 500)}
```
adbindal, line 55: I'd refrain from using magic numbers in the invocation. Instead, please consider creating a const variable and using that directly inside the function. That should also help refactor this code as: -
```suggestion
        onGetErrorMessage={LinkTextValidation}
```

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -34,12 +52,15 @@ const AssignmentLinkEditorInner = ({
...
         styles={themedClassNames(textFieldBaseStyle)}
         onChange={(_e, newValue) => onChangeTextField('displayText', newValue || '')}
+        onGetErrorMessage={(value) => LinkTextValidation(value, 500)}
       />
       <TextField
         value={link?.url || ''}
         placeholder="URL"
         styles={themedClassNames(urlInputStyle)}
         onChange={(_e, newValue) => onChangeTextField('url', newValue || '')}
+        onGetErrorMessage={(value) => isValidURL(value)? '' : 'Invalid URL'}
```
adbindal, line 62: I'm a bit confused as to why we've returned string for `TextLengthValidation` and not for `UrlValidation`? We could consider following the same practice for both cases. You can wrap `isValidUrl` inside the function to help this transition.

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -18,6 +18,24 @@ interface AssignmentLinkEditorProps {
 }
 export type AssignmentLinkEditorStyles = SimpleComponentStyles<'root'>;
 
+const LinkTextValidation = (value: string | undefined, MAXLEN: number) => {
```
adbindal, line 21: Seems like you wish to re-use this function mutliple times. Probably you can use currying to derive 2 different functions out of this to help make the intent much cleaner on this front.
``` javascript
const ValidateTextLength = (limit: number) => (text: string | undefined) => (text?.length < limit) : "Text entered is too long: '';
const ValidateTitleLength = ValidateTextLength(500);
const ValidateDescLength = ValidateDescLength(1000);
```

client/src/Features/AssignmentLinks/useLink.hook.ts
```diff
@@ -5,15 +5,21 @@
...
+import { isValidURL } from './AssignmentLinkEditor';
 
 type useLinkReturnType = [AssignmentLink, (link: AssignmentLink) => void, boolean];
 
 export const useLink = (assignmentLink: AssignmentLink): useLinkReturnType => {
   const [link, setLink] = useState<AssignmentLink>(assignmentLink);
-  const [canSave, setCanSave] = useState(!!link.url);
+
+  const isValidLink = (link: AssignmentLink) => {
+    return link.url.length>0 && isValidURL(link.url) && link.displayText.length<=500 && link.description.length<=1000;
```
adbindal, line 16: Reusing the magic numbers here again is usually not scalable. We may end up updating these numbers at on place leaving the other. How about refactoring Length check function for reusability here?

adbindal: Shouldn't `isValidUrl` take care of `link.url.length > 0` check?

adbindal: For me, it would be amazing if the function could read something like below.
``` javascript
const isValidLink = ({description, displayText, url}: AssignmentLink) => isValidTitle(displayText) && isValidDesc(description) && isValidUrl(url);
```

adbindal (commented):
Thanks for the changes. I hope similar checks would be added on the backend as well. Doing hardening on backend is of utmost importance as that code is not under client control.

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
maleekha: I have added some refactoring changes for this, as well as for the other comments.

client/src/Features/AssignmentLinks/useLink.hook.ts
maleekha: I have added some changes to refactor this portion.

Wrt `Shouldn't isValidUrl take care of link.url.length > 0 check?` : When the link is undefined or empty, we don't want to show the "Invalid message", but we want it to disable the `Add`/`Save` buttons. For this reason, I had added this condition here instead of inside `isValidLink`.

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
maleekha: It actually gives an error, I tried this on my end!

client/src/Features/AssignmentLinks/useLink.hook.ts
adbindal: Makes sense. However, in that case we should not even be calling `isUrlValid` or `isValidLink` -- right?

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -18,6 +18,22 @@ interface AssignmentLinkEditorProps {
 }
 export type AssignmentLinkEditorStyles = SimpleComponentStyles<'root'>;
 
+const ValidateTextLength = (limit: number) => (text: string | undefined) => (text && text?.length < limit) ? "The entered text is too long." : "";
+export const ValidateTitleLength = ValidateTextLength(500);
+export const ValidateDescLength = ValidateTextLength(1000);
+
+export const isValidURL = (value: string | undefined) => {
```
adbindal, line 25: Should we also check for cases that the url is valid **HTTPUrl** 🤔 ?

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
```diff
@@ -34,12 +50,15 @@ const AssignmentLinkEditorInner = ({
...
         styles={themedClassNames(textFieldBaseStyle)}
         onChange={(_e, newValue) => onChangeTextField('displayText', newValue || '')}
+        onGetErrorMessage={ValidateTitleLength}
       />
       <TextField
         value={link?.url || ''}
         placeholder="URL"
         styles={themedClassNames(urlInputStyle)}
         onChange={(_e, newValue) => onChangeTextField('url', newValue || '')}
+        onGetErrorMessage={ValidateURL}
```
adbindal, line 60: Just wondering if there's a way to simulate the following inside `<TextField {...props} />`

> **Reference**: https://www.w3schools.com/tags/att_input_type_url.asp

client/src/Features/AssignmentLinks/useLink.hook.ts
```diff
@@ -5,16 +5,24 @@
...
 import { useState, useEffect } from 'react';
+import { isValidURL, ValidateDescLength, ValidateTitleLength } from './AssignmentLinkEditor';
 
 type useLinkReturnType = [AssignmentLink, (link: AssignmentLink) => void, boolean];
 
 export const useLink = (assignmentLink: AssignmentLink): useLinkReturnType => {
   const [link, setLink] = useState<AssignmentLink>(assignmentLink);
-  const [canSave, setCanSave] = useState(!!link.url);
+
+  const titleIsValid = ValidateTitleLength(link.displayText) === "";
```
adbindal, line 15: The unfortunate part of handling it this way is that the implementation calls all 3 functions before deciding if the link is valid or not. In previous case, that was not required if any of the checks failed.

> Essentially, I liked the previous approach :).

adbindal (changes requested):
Overall, looks good to me. Have added some not so important comments. One important thing, however, is its backend counterpart. I should be good to go as soon as you could point me to it (or a follow-up work item to take care of it).

maleekha:
> Overall, looks good to me. Have added some not so important comments. One important thing, however, is its backend counterpart. I should be good to go as soon as you could point me to it (or a follow-up work item to take care of it).

I was under the impression that the client-side validation would be sufficient, but based on your previous comment as well, I see why it might be more important. I have pushed a change now, wherein we make these checks on the incoming dto as well.

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
maleekha: Added a check for the `url.protocol` to the client, and the backend.

client/src/Features/AssignmentLinks/useLink.hook.ts
maleekha: I have made the changes, that should help here!

client/src/Features/AssignmentLinks/AssignmentLinkEditor.tsx
maleekha: I began exploring this itself before beginning the task if we could find some way to avoid writing the validation checks by simply adding a type field as `url`. Unfortunately, couldn't find a parallel property that could serve the purpose for the `TextField`.

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
```diff
@@ -109,6 +109,16 @@ public AssignmentLinksApi(ILogger<AssignmentLinksApi> logger, IMapper mapper)
             if (linkId != linkDto.Id)
                 return new BadRequestErrorMessageResult("The provided link content doesn't match the path.");
 
+            Uri linkUri;
```
adbindal, line 112: I'd recommend using DataAnnotations for C#. It would be much cleaner and scalable for future.

> **Reference**: https://docs.microsoft.com/en-us/aspnet/mvc/overview/older-versions/getting-started-with-aspnet-mvc3/cs/adding-validation-to-the-model

adbindal (changes requested):
Sorry for another round. Have added some comments inline.

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
maleekha: Thank you for this suggestion. I have made the necessary changes to use data annotations and validation.

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinkEntity.cs
```diff
@@ -4,13 +4,20 @@
 // --------------------------------------------------------------------------------------------
 
 using Microsoft.WindowsAzure.Storage.Table;
+using System.ComponentModel.DataAnnotations;
 
 namespace Edna.AssignmentLinks
 {
     public class AssignmentLinkEntity : TableEntity
     {
+        [StringLength(500)]
```
adbindal, line 13: 🎉

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
```diff
@@ -115,6 +116,13 @@ public AssignmentLinksApi(ILogger<AssignmentLinksApi> logger, IMapper mapper)
             assignmentLinkEntity.PartitionKey = assignmentId;
             assignmentLinkEntity.ETag = "*";
 
+            System.ComponentModel.DataAnnotations.ValidationContext context = new System.ComponentModel.DataAnnotations.ValidationContext(assignmentLinkEntity, null, null);
+            bool isValid = Validator.TryValidateObject(assignmentLinkEntity, context, new List<ValidationResult>(), true);
+            if (!isValid)
+            {
+                return new BadRequestErrorMessageResult("The provided link field values are not valid.");
+            }
```
adbindal, line 124:
```suggestion dotnet
            ValidationContext context = new ValidationContext(assignmentLinkEntity, null, null);
            if (!Validator.TryValidateObject(assignmentLinkEntity, context, new List<ValidationResult>(), true))
            {
                return new BadRequestErrorMessageResult("The provided link field values are not valid.");
            }
```

adbindal (approved):
Minor comments inline. Looks Good now 👍

PR #92, opened 2020-11-02
Enable auditing on azure key vault

nehalilani:
### Additions
Enabling auditing on the key Vault with a retention period of 90 days.


deployment/azuredeploy.json
```diff
@@ -49,6 +49,7 @@
         "staticWebsite": "[toLower(concat('learnclient', replace(variables('uniqueIdentifier'),'-','')))]",
         "keyVaultName": "[toLower(concat('kv-', variables('uniqueIdentifier')))]",
         "StorageAccountSuffix": "[toLower(concat(variables('uniqueIdentifier')))]",
+        "AuditLogsStorageAccountName": "[toLower(concat('auditlogs' , variables('StorageAccountSuffix')))]",
```
adbindal, line 52: Should we somehow indicate that these are KeyVault Audit logs 🤔 ?
> Maybe we can consider using `KeyVaultName` as second param of concat instead of `StorageAccountSuffix`, so that the 2 gets correlated.

deployment/azuredeploy.json
```diff
@@ -416,6 +426,25 @@
                 }
             }
         },
+        {
+            "type": "Microsoft.KeyVault/vaults/providers/diagnosticSettings",
+            "apiVersion": "2017-05-01-preview",
+            "name": "[concat(variables('keyVaultName'), '/Microsoft.Insights/Send to all locations')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.KeyVault/vaults', variables('keyVaultName'))]"
```
adbindal, line 434: Why are we not depending it on `AuditLogStorageAccountName` as well?

adbindal (changes requested):
The changes look fine to me. Added a few thoughts related to naming and dependency resolution. Once cleared, these are good to be merged.

deployment/azuredeploy.json
nehalilani: seems like that got missed. thanks for pointing it out. adding it in the next commit.

deployment/azuredeploy.json
```diff
@@ -49,6 +49,7 @@
         "staticWebsite": "[toLower(concat('learnclient', replace(variables('uniqueIdentifier'),'-','')))]",
         "keyVaultName": "[toLower(concat('kv-', variables('uniqueIdentifier')))]",
         "StorageAccountSuffix": "[toLower(concat(variables('uniqueIdentifier')))]",
+        "KeyVaultAuditLogsStorageAccountName": "[toLower(concat('auditlogskv' , variables('StorageAccountSuffix')))]",
```
adbindal, line 52: > What I meant from my previous comment was something like below. This should group KV resources together on the UI as well. Let me know if this seems better: -
```suggestion json
        "KeyVaultAuditLogsStorageAccountName": "[toLower(concat(variables('keyVaultName'), '-auditlogs' ))]",
```

adbindal (approved):
Changes look good now. One minor name-change request, if looks fine.

deployment/azuredeploy.json
nehalilani: storage account names cannot include special characters, so basically the storage account suffix is the unique identifier without the `-`.
Right now the keyvault name is `kv-<storageAccountSuffix>` and according to this code, the storage account name for logging is `auditlogskv<storageAccountSuffix>`

Having said that, i'll change it to `kv<storageAccountSuffix>auditlogs`

adbindal (approved):

PR #94, opened 2020-11-10
Adding caution msg to clean up script

nehalilani:
### Changes - 
Added a warning message in the cleanup script, to use a different Resource Group name as the soft-delete option for Key Vault is enabled and another key vault with the same name cannot be created.


deployment/Cleanup.ps1
```diff
@@ -174,5 +174,6 @@ process {
     Write-Title '======== Successfully Deleted Resources from Azure ==========='
 
     Write-Log -Message "Clean-up Complete"
+    Write-Host "Please use a different ResourceGroup name on re-deployment!"
```
adbindal, line 177: How about using [Write-Warning](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-warning?view=powershell-7) instead with similar message 🤔?

adbindal (changes requested):

adbindal (approved):

PR #95, opened 2020-11-10
Enabling Logs for Static Website

maleekha:
**Changes made**
Added an azure-cli command to enable read-write-delete logs for the static website.

**Testing**
- Deployment testing: to ensure that the script does not fail
- Deleted a file from the container to ensure that the logs are added.


deployment/Install-Client.ps1
```diff
@@ -162,6 +162,8 @@ function Install-Client {
         Write-ClientDebugLog -Message 'Uploading build content to the `$web storage container'
         # Turning Error only mode as this cmd shows a warning which causes misconception that user needs to sign in to azure cli
         $result = az storage blob upload-batch -s 'build' -d '$web' --account-name $StaticWebsiteStorageAccount --only-show-errors | ConvertFrom-Json
+        az storage logging update --log rwd --retention 30 --services b --account-name $StaticWebsiteStorageAccount --only-show-errors
```
adbindal, line 165: Question: What happens in case this fails? Should we try adding some error handling here?

adbindal (changes requested):

deployment/Install-Client.ps1
maleekha: I agree, I have updated the change. Thanks for pointing.

deployment/Install-Client.ps1
```diff
@@ -162,6 +162,14 @@ function Install-Client {
         Write-ClientDebugLog -Message 'Uploading build content to the `$web storage container'
         # Turning Error only mode as this cmd shows a warning which causes misconception that user needs to sign in to azure cli
         $result = az storage blob upload-batch -s 'build' -d '$web' --account-name $StaticWebsiteStorageAccount --only-show-errors | ConvertFrom-Json
+        
+        try{
+            az storage logging update --log rwd --retention 30 --services b --account-name $StaticWebsiteStorageAccount --only-show-errors
+        } 
+        catch {
+            Write-Output "Could not enable logging for static website storage account"
```
adbindal, line 170: Since we've handled this as a non-fatal error (and it is indeed non-fatal to deployment), few things coming to my mind: -
* Should we log it as Warning instead of a normal `Write-Output`?
* Any reason for omitting the `Write-ClientDebugLog` to log the complete exception?
* Could we consider writing a small doc (FAQ maybe) around how to do this manually and point the aka.ms url of that document here 🤔?
* Maybe rephrase the Warning to be more like "_Failed enable auditing for static website. This could be a security risk, please enable auditing for static website manually via `aka.ms/learn-lti-audit-client`_"

adbindal (changes requested):
Minor comments inline on the newer changes.

deployment/Install-Client.ps1
```diff
@@ -162,6 +162,14 @@ function Install-Client {
         Write-ClientDebugLog -Message 'Uploading build content to the `$web storage container'
         # Turning Error only mode as this cmd shows a warning which causes misconception that user needs to sign in to azure cli
         $result = az storage blob upload-batch -s 'build' -d '$web' --account-name $StaticWebsiteStorageAccount --only-show-errors | ConvertFrom-Json
+
```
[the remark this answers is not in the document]
maleekha: I double-checked: Turns out the command does not return anything even when it fails. When it fails, it logs the failure message to the command line but does not return anything per se. I have modified the error handling for this command to check the $LASTEXITCODE value to check for errors (non-zero would mean error). Any other way you suggest we could take this up?

deployment/Install-Client.ps1
```diff
@@ -162,6 +162,14 @@ function Install-Client {
...
         $result = az storage blob upload-batch -s 'build' -d '$web' --account-name $StaticWebsiteStorageAccount --only-show-errors | ConvertFrom-Json
+        
+        try{
+            az storage logging update --log rwd --retention 30 --services b --account-name $StaticWebsiteStorageAccount --only-show-errors
+        } 
+        catch {
+            Write-Output "Could not enable logging for static website storage account"
+        }
+
         if(!$result) {
```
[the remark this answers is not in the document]
maleekha: Enabling it before we upload the content would help to save all the logs for that container, right from the start. Can move it up in that case!

deployment/Install-Client.ps1
maleekha: I have created a link as per your suggestion. I will update the content shortly to include the manual steps for enabling auditing. I will update the error message shortly!

maleekha: I have added the link to the manual steps mentioned in the docs.microsoft guide for now. I hope that works. Can make it a part of our documentation as a follow-up.

deployment/Install-Client.ps1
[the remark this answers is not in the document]
maleekha: Checking the $LASTEXITCODE is being done at a couple of other places in our code as well. I believe that shouldn't be a problem. I have added the error handling using the same.

maleekha:
I have addressed the comments here and tested the changes as well on my end.

client/.env.production
```diff
@@ -1,10 +1,10 @@
 GENERATE_SOURCEMAP='false'
-REACT_APP_EDNA_PLATFORM_SERVICE_URL='https://edna-lti-ab-platfomrs.azurewebsites.net/api'
-REACT_APP_EDNA_USERS_SERVICE_URL='https://edna-lti-ab-users.azurewebsites.net/api'
-REACT_APP_EDNA_LEARN_CONTENT='https://edna-lti-ab-learncontent.azurewebsites.net/api'
-REACT_APP_EDNA_LINKS_SERVICE_URL='https://edna-lti-ab-links.azurewebsites.net/api'
-REACT_APP_EDNA_ASSIGNMENT_SERVICE_URL='https://edna-lti-ab-assignments.azurewebsites.net/api'
+REACT_APP_EDNA_PLATFORM_SERVICE_URL='https://platforms-crbr22nms.azurewebsites.net/api'
```
adbindal, line 2: Any reason for making these changes?

deployment/Install-Client.ps1
```diff
@@ -152,6 +152,15 @@ function Install-Client {
             throw 'Errors while creating Optimized Production Build'
         }
 
+        az storage logging update --log rwd --retention 30 --services b --account-name $StaticWebsiteStorageAccount --only-show-errors
+        # Checking if the command for enabling logs of storage account failed
+        if($LASTEXITCODE -ne 0){
+            $manualStepsURL = "https://docs.microsoft.com/en-us/azure/storage/common/storage-monitor-storage-account#configure-logging"
```
adbindal, line 158: I believe we've created an aka.ms link for the same. We might consider using that here.

adbindal (changes requested):
Changes look good. Minor changes suggested inline.

client/.env.production
maleekha: Checked this in by mistake. Reverted the change.

deployment/Install-Client.ps1
maleekha: Done. I had created the link, yes, but I was under the impression we could host a separate link via `aka.ms` by modifying the content somewhere; just realized that it is primarily a redirection service. I have updated the link now.

adbindal (approved):

PR #97, opened 2020-11-11
Assignment Link Validation for LinkIDs

maleekha:
**Changes Made**

- Changed the linkdto id datatype from `string` to `Guid`
- Added error handling in the SAVE and GET Api calls for cases where the incoming request or the storage contains a malformed linkID which cannot be parsed to `Guid`
- Added UI changes for when certain linkIDs are invalid and are not being rendered


backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
```diff
@@ -104,9 +107,16 @@ public AssignmentLinksApi(ILogger<AssignmentLinksApi> logger, IMapper mapper)
...
+            try
+            {
+                linkDto = JsonConvert.DeserializeObject<AssignmentLinkDto>(linkJson);
+            } catch (Exception ex)
+            {
+                return new BadRequestErrorMessageResult("Could not create a valid link from the request. " + ex.Message);
+            }
 
-            if (linkId != linkDto.Id)
+            if (linkId != linkDto.Id.ToString())
```
adbindal, line 119: What would it take to start accepting `linkId` as **Guid** also?

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
```diff
@@ -62,6 +62,7 @@ public AssignmentLinksApi(ILogger<AssignmentLinksApi> logger, IMapper mapper)
 
             IEnumerable<AssignmentLinkDto> assignmentLinkDtos = sameAssignmentLinks
                 .OrderBy(linkEntity => linkEntity.Timestamp.Ticks)
+                .Where(linkEntity => Guid.TryParse(linkEntity.RowKey, out Guid result))
```
adbindal, line 65: Does it make sense to move this into `_mapper`? That way it should help in case newer entry points for LinkEntity -> LinkDto conversion popup in future.

adbindal (changes requested):
The changes seem fine with me. I've shared some queries and thoughts relating to moving the fixes to a more common location.

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
maleekha: The row key in an azure table can only be a string; we cannot override this explicitly, which is why we are having to put this validation in the dto.

backend/Functions/Edna.AssignmentLinks/Edna.AssignmentLinks/AssignmentLinksApi.cs
maleekha: I have made some changes in this regard: essentially, the mapper while converting `LinkEntity` to `LinkDto` will map the Id only if it can be parsed to a `Guid`, else it will keep it as the default `Guid.Empty` value. 

On the client-side, I have added a check for filtering such links with `Guid.Empty` Ids, indicating that these are not "well-formed". Although the same holds here also: the check might have to be propagated in the client-side if a similar condition arises somewhere else. But given the store can be re-used where ever we want, in future, it is less likely we will have to rewrite such conditions. Any thoughts?

maleekha:
I have made some changes to address the comments. In particular, I have moved the invalid-link-filtering part to the `link store`, and have also added some UI changes to reflect whenever certain links are not being rendered due to invalid-ids.

adbindal (approved):
LGTM 👍

PR #98, opened 2020-11-12
Validation for string inputs in Assignment

nehalilani:
## Changes

- Assignment Name & Course name: max length = 100 characters
- Assignment Description: max length = 2500 characters
- LTI version: convert to Enum
- Publish status: convert to Enum
- Context Membership URL: add URL validation

### Testing
Completed


backend/Functions/Edna.Connect/Lti1Api.cs
```diff
@@ -78,10 +78,23 @@ private Assignment ConvertRequestToAssignment(LtiRequest ltiRequest)
...
                 Name = ltiRequest.ResourceLinkTitle,
                 CourseName = ltiRequest.ContextTitle,
-                LtiVersion = ltiRequest.Version,
+                LtiVersion = GetLtiVersion(ltiRequest),
                 ContextMembershipsUrl = ltiRequest.Parameters.FirstOrDefault(pair => pair.Key == "custom_context_memberships_url").Value,
                 OAuthConsumerKey = ltiRequest.ConsumerKey
             };
         }
+
+        private static string GetLtiVersion(LtiRequest ltiRequest)
```
adbindal, line 87: Wondering if we should be moving this to `LtiVersion` Enum?

backend/Functions/Edna.Connect/Lti1Api.cs
```diff
@@ -78,10 +78,23 @@ private Assignment ConvertRequestToAssignment(LtiRequest ltiRequest)
...
+        {
+            switch(ltiRequest.Version)
+            {
+                case "1.0":
+                    return LtiVersion.Lti1.ToString();
+                case "1.3.0":
+                    return LtiVersion.LtiAdvantage.ToString();
+                default:
+                    return null;
```
adbindal, line 96: Should we log this information so that we at least know what is getting passed to us at this location 🤔 ?

backend/Functions/Edna.Connect/LtiAdvantageApi.cs
```diff
@@ -106,6 +106,9 @@ public async Task SaveState([OrchestrationTrigger] IDurableOrchestrationContext
             await assignmentsCollector.AddAsync(assignment);
             await assignmentsCollector.FlushAsync();
 
+            if (string.IsNullOrEmpty(assignment.Id))
```
adbindal, line 109: Amazing. Let's add a descriptive message / log to indicate that the failure is due to missing assignment id.

client/src/Features/GeneralPage/DescriptionInput.tsx
```diff
@@ -13,6 +13,9 @@ import { textFieldBaseStyle } from '../../Core/Components/Common/Inputs/EdnaInpu
 import { useStore } from '../../Stores/Core';
 
 type DescriptionInputStyles = Partial<ITextFieldStyles>;
+
+const ValidateDescLength = (text: string | undefined) => (text && text?.length <= 2500) || !text ? "" : "The entered text is too long.";
```
adbindal, line 17:
```suggestion
const ValidateDescLength = (text: string | undefined) => text ? (text.length <= 2500) ? text :  "The entered text is too long." : "";

```

adbindal (commented):
Changes look good. Minor comments inline.

backend/Functions/Edna.Connect/LtiAdvantageApi.cs
nehalilani: assignment Id is not created from context id and resource link id when the assignment doesn't save successfully, an error message with the reason for failure does get logged.

backend/Functions/Edna.Connect/Lti1Api.cs
nehalilani: LtiRequest for LTI1 and LTI1.3 is different, so moving this function to the enum won't be useful

adbindal: Apologies, I don't understand. Are we saying that string values are different for mapping to the same enum value? or is it something else?

backend/Functions/Edna.Connect/LtiAdvantageApi.cs
adbindal: Again, I am unable to understand your answer. From where does the logging happens in the above case? Are we seeing this in the logs where this is happening? Request you to please point me to the location to help me get the necessary context.

backend/Functions/Edna.Connect/Lti1Api.cs
nehalilani: the request type is different for lti 1 and lti 1.3
For LTI1, the request is of the type `LtiLibrary.NetCore.Lti.v1.LtiRequest` whereas for LTI1.3, the request type is `System.IdentityModel.Tokens.Jwt.JwtPayload`
Moving this function to the enum would not be helpful as separate functions would need to be created there as well

backend/Functions/Edna.Connect/LtiAdvantageApi.cs
nehalilani: Assignment.Id is gets updated in the SaveAssignment function (which is called from the flushAsync function) when the assignment gets saved successfully.
In case the assignment save is not successful, assignment.Id is not updated (remains null) and an error msg with appropriate reason gets logged in the LtiAdvantage/ Lti1 function.
The logging happens from the SaveAssignment function in the `AssigmentsClient.cs` file.

I hope this clears your doubt.

nehalilani: The same check is present in the Lti1Api.cs file as well.

client/src/Features/StudentView/StudentViewContent.tsx
```diff
@@ -37,7 +37,7 @@ const StudentViewContentInner = ({ styles, requirePublished }: StudentViewConten
     const items: (StudentViewSectionProps & IStylesOnly<StudentViewSectionStyles>)[] = _.compact([
       assignmentStore.assignment?.description && {
         title: 'Description',
-        textContent: assignmentStore.assignment.description
+        textContent: assignmentStore.assignment.description.substring(0,2500)
```
adbindal, line 40: Follow-up: we could consider `substring`ing to 2497 and append `...` to indicate that the description has been trimmed.

adbindal (approved):
LGTM 👍. Thanks for the changes.

Issue #96, opened 2020-11-11
Course completion tracking of Moodle Integration


koolala:
Some problem/questions about the course completion of Moodle Integration. 

Moodle Version: 3.8
MS Learn LTI Application: Should be latest (clone and re-deploy on 9 Sept, 2020)

**Describe the issue**
1. Launch Container
If set to open in New Window, the learner's activity never auto change to completed.
![image](https://user-images.githubusercontent.com/3178741/98778616-995bf180-242d-11eb-9f4a-925d6e922907.png)

if set to Embed, the the learner's activity can auto change to completed. After the learner open the embed ms learn application page either he/she not click the ms learn modules or not. 

2. Completion Tracking
In above reason, we try to enabled the course completion tracking. 
![image](https://user-images.githubusercontent.com/3178741/98778175-eab7b100-242c-11eb-8d0e-6edd99e3cbd9.png)
- Require Guide: Enabled

But the completion status still cannot change to completed.

**Expected behavior**
Expect the completion tracking should receive the grade from MS Learn LTI Application, Moodle Course according the grade to update learner's activity status of the course.
![image](https://user-images.githubusercontent.com/3178741/98782974-ca3f2500-2433-11eb-8339-7b837d323307.png)

nehalilani:
Hi @koolala ,

Thanks for reporting this issue, to answer your questions - 

1. It is advisable to set the Launch Container, to open in New Window.

2. For Completion Tracking, this functionality is not supported yet, hence this is the expected behavior.

leestott:
@koolala today we launched 

Organizational Reporting as  service available to organizations to view Microsoft Learn training progress and achievements of the individuals within their tenant. This service is available to both enterprise customers and educational organizations.

The system uses a service called Azure Data Share to extract, transform, and load (ETL) user progress data into data sets, which can then be processed further or displayed in visualization tools such as Power BI. Data sets can be stored to either Azure Data Lake, Azure Blob storage, Azure SQL database, or Azure Synapse SQL Pool.

Organizations can create and manage their data share using Azure Data Share’s no-code UI. https://docs.microsoft.com/en-us/learn/support/org-reporting

leestott:
@koolala did the Microsoft Learn Organisational reporting resolve your issue of understanding learners usage of the LTI application and Microsoft Learn resources?

Can we close this issue?

PR #104, opened 2020-12-01
Bumped y18n package version to 5.0.5

nehalilani:
for reference-
issue: https://github.com/yargs/y18n/issues/96
fix: https://github.com/yargs/y18n/pull/108


client/package.json
```diff
@@ -26,7 +26,8 @@
     "react-router-dom": "^5.1.2",
     "react-virtualized-auto-sizer": "^1.0.2",
     "react-window": "^1.8.5",
-    "uuid": "^8.0.0"
+    "uuid": "^8.0.0",
+    "y18n": "^5.0.5"
```
adbindal, line 30: Why are we adding this dependency directly inside `package.json`? `y18n` should be coming to us via some transitive relationship. IMHO, let's try to understand what that relation is and then try fixing the dependency that we actually depend on, instead of having adding a dependency on a new component.

adbindal (changes requested):

client/package.json
nehalilani: on installing the latest version of the y18n package, this dependency got added directly in package.json 
sure, I'll see what the relation is for y18n

adbindal (approved):

nehalilani:
This is a transitive dependency and shouldn't be updated like this.

Muhammadariesta (approved):

PR #151, opened 2021-08-04
Add an error string when user is missing a name, prevents infinite spinner

adbindal:
## Background
Fixes the blocker faced in **OpenEdx LTI Integration**, there the user does not have a name and the experience shows an un-ending `Loading` spinner on the screen.

@Lee - Would you please help with the appropriate string here? The case is that we would like the user to have a name to be shown on the screen. Without that, I've added an error instead for now.


leestott (approved):

leestott:
Updating to provide error message if the user has no name 
  this.errorContent = { errorMsg: 'Please add a name for the user.', icon: 'BlockContact' };

PR #156, opened 2021-08-27
Adding OAuth Audience field for Brightspace D2L support

alphaz18:
Adding customizable Audience field to be passed around for token. 
Originally, the aud claim in the token was automatically set to the tokenEndpointUrl. 
However, per (https://www.imsglobal.org/spec/security/v1p0) the audience claim MAY be the token endpoint url but is not necessarily that. 
Some lms platforms set a different aud claim, and thus would fail using LTI-Learn. Eg. Brightspace. 

I made the field optional, so as to hopefully not break existing deployments, if blank, it should use the token endpoint as per previous. I haven't tested on Moodle, canvas or blackboard since I don't those platforms, only Brightspace with LTI 1.3.


leestott:
@alphaz18 thanks for PR regarding the outh support for Brightspace D2L #156

Could you also help us develop setup/config guides for D2L and PR changes for https://github.com/microsoft/Learn-LTI/blob/main/docs/CONFIGURATION_GUIDE.md to include D2L see https://github.com/microsoft/Learn-LTI/issues/68

alphaz18:
@leestott Hi, I just pushed changes to the PR for the guide as requested, Hopefully that helps, Regards,

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/AccessTokenService.cs
```diff
@@ -38,7 +38,7 @@ public async Task<TokenResponse> GetAccessTokenAsync(string clientId, string acc
             var payload = new JwtPayload();
             payload.AddClaim(new Claim(JwtRegisteredClaimNames.Iss, clientId));
             payload.AddClaim(new Claim(JwtRegisteredClaimNames.Sub, clientId));
-            payload.AddClaim(new Claim(JwtRegisteredClaimNames.Aud, accessTokenEndpoint));
```
adbindal, line 41: Is it possible to perform this ternary check in the platform itself? Passing it all over across all these functions could be avoided if we could update the platform to store correct Aud itself.

adbindal (commented):
Thanks a lot for getting these in. I have only one doubt on them which I've placed inline, rest of it looks ok.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/AccessTokenService.cs
alphaz18: hi @adbindal, I would have been happy to, if you could explain how I could do it. the platform is not being passed to the getaccesstokenasync. 
So the way I saw it, even if I moved the check to the platform, I would still have to pass the audience all over the place, because how else do you get the audience to the method? eg, look at the method declaration, it needs client id, accesstokenendpoint being passed all over the place..

adbindal: Sure. I've kind of lost track of the code in my mind so unable to talk too much in terms of functions and files to change. The aim was to modify the Platform Model to store the Audience correctly instead of storing both AccessTokenUrl & Audience. This way we would only be passing `audience` param to the `AccessTokenService`. I'd go ahead and re-read the code to remember those linkages in the code, in order to help you better.

alphaz18: I think the issue here is you actually need both parameters 
accesstokenendpoint is necessary as the target address of where to request token from, like on line 51 of that file. 
And audience is necessary for jwtRegisteredClaimNames.Aud. 

I would almost consider it a bug, that the accessTokenService was using the accesstokenendpoint for the JwtRegisteredClaimsName.Aud. in my original post, I mentioned essentially that
The imsglobal spec basically mentioned that it's common to use the tokenendpoint as the Audience, but that it's not a given and it may use any other value. in the implementation here previously, it was basically hard coded to have thte audience as the tokenEndpoint. so while some LMS' like blackboard or Moodle, will send the endpoint as the audience by default, others like Brightspace cannot use the same value.

client/src/Features/Platform/PlatformRegistrationFields.ts
```diff
@@ -78,6 +78,13 @@ const toolSettingsFields: field[] = [
     isRequired: true,
     isCopyable: false
   },
+  {
+    fieldLabel: 'Audience',
+    fieldName: 'audience',
+    helpText: 'The Audience provided by the LMS for this specific registration. If Blank will use the endpoint URL.',
```
nehalilani, line 84: @leestott Can you please review the helpText here once?

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/AccessTokenService.cs
nehalilani: The current changes looks fine for now

nehalilani: Later on, we can start passing the platform object

client/src/Features/Platform/PlatformRegistrationFields.ts
leestott: yes this looks fine. We will add the details of Audience field to OpenEdx, Canvas, Moodle and Blackboard as NOT being applicable so can be left blank.

Regression testing from an engineering perspective to ensure changes do not impact other LMS installations

PR #171, opened 2022-01-28
Bugfix for more than 50 course members not visible

maitreeleekha:
> Using the "next" link header from NRPS response to fetch the paginated member response, as per LTI spec: https://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter

Tested in Canvas, Blackboard & Moodle 
Limit only hitting in Canvas

### DIFF
--- a/backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
+++ b/backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
@@ -15,6 +15,7 @@ using LtiAdvantage.NamesRoleProvisioningService;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Edna.Utils.Http;
+using System.Text.RegularExpressions;
 
 namespace Edna.Bindings.LtiAdvantage.Services
 {
@@ -51,7 +52,7 @@ namespace Edna.Bindings.LtiAdvantage.Services
             _accessTokenService = accessTokenService;
             _logger = logger;
         }
-        
+
         public async Task<IEnumerable<Member>> Get(string clientId, string tokenUrl, string audience, string membershipUrl)
         {
             _logger.LogInformation("Getting token for client ID: " + clientId);
@@ -70,18 +71,27 @@ namespace Edna.Bindings.LtiAdvantage.Services
             httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(Constants.MediaTypes.MembershipContainer));
 
             _logger.LogInformation("Getting members.");
-            using var response = await httpClient.GetAsync(membershipUrl);
+            List<Member> members = new List<Member>();
 
-            if (!response.IsSuccessStatusCode)
+            do
             {
-                _logger.LogError("Could not get members.");
-                throw new Exception(response.ReasonPhrase);
-            }
+                using var response = await httpClient.GetAsync(membershipUrl);
 
-            var content = await response.Content.ReadAsStringAsync();
-            var membership = JsonConvert.DeserializeObject<MembershipContainer>(content);
-            
-            return membership.Members
+                if (!response.IsSuccessStatusCode)
+                {
+                    _logger.LogError("Could not get members.");
+                    throw new Exception(response.ReasonPhrase);
+                }
+
+                var content = await response.Content.ReadAsStringAsync();
+                var membership = JsonConvert.DeserializeObject<MembershipContainer>(content);
+                members.AddRange(membership.Members);
+
+                membershipUrl = GetNextMembershipUrlFromHeaders(response.Headers);
+            }
+            while (!string.IsNullOrEmpty(membershipUrl));
+
+            return members
                 .OrderBy(m => m.FamilyName)
                 .ThenBy(m => m.GivenName);
         }
@@ -91,17 +101,18 @@ namespace Edna.Bindings.LtiAdvantage.Services
             // Looks like LTI 1.3 doesn't support querying by member identifiers
 
             IEnumerable<Member> allMembers = await Get(clientId, tokenUrl, audience, membershipUrl);
-            foreach (Member m in allMembers) {       
+            foreach (Member m in allMembers)
+            {
                 string name = m.Email;
-                if (m.Email != null && ( m.FamilyName == null || m.GivenName == null))
-                {            
+                if (m.Email != null && (m.FamilyName == null || m.GivenName == null))
+                {
                     int index = name.IndexOf("@");
                     name = name.Substring(0, index);
                     m.FamilyName = name;
                     m.GivenName = " ";
                 }
             }
-            return allMembers.FirstOrDefault(member => userEmails.Any(userEmail => (member.Email??String.Empty).Equals(userEmail, StringComparison.OrdinalIgnoreCase)));
+            return allMembers.FirstOrDefault(member => userEmails.Any(userEmail => (member.Email ?? String.Empty).Equals(userEmail, StringComparison.OrdinalIgnoreCase)));
         }
 
         public async Task<Member> GetById(string clientId, string tokenUrl, string audience, string membershipUrl, string userId)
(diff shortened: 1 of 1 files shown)


leestott (approved):

maitreeleekha:
Tested the changes in Canvas; Pending in Blackboard and Moodle

leestott:
> Tested the changes in Canvas; Pending in Blackboard and Moodle

Thanks for the confirmation. I have added you as a reviewer to approve after testing.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
```diff
@@ -70,18 +71,27 @@ public async Task<IEnumerable<Member>> Get(string clientId, string tokenUrl, str
             httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(Constants.MediaTypes.MembershipContainer));
 
             _logger.LogInformation("Getting members.");
-            using var response = await httpClient.GetAsync(membershipUrl);
+            List<Member> members = new List<Member>();
 
-            if (!response.IsSuccessStatusCode)
+            do
```
adbindal, line 76: Quick Question on the `do..while`, it looks like we are proactively fetching all the results and sending them over to the client here. I was wondering if this could be done lazily to allow the client to call the next `membershipurl` as and when needed. Does that mean changing a lot of functionality on the client - such as `filter`, `sortBy` and `search`? If not, probably we can try to use that approach to help reduce user bandwidth for these cases 🤔 ...

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
```diff
@@ -111,5 +122,32 @@ public async Task<Member> GetById(string clientId, string tokenUrl, string audie
...
+        private static string GetNextMembershipUrlFromHeaders(HttpResponseHeaders headers)
+        {
+            string linkUrls = headers.TryGetValues("Link", out var values) ? values.FirstOrDefault() : null;
+            if (string.IsNullOrEmpty(linkUrls))
+            {
+                return null;
+            }
+
+            string[] urls = linkUrls.Split(',');
+            string nextUrl = urls.FirstOrDefault(u => u.IndexOf("\"next\"", StringComparison.OrdinalIgnoreCase) >=0 );
```
adbindal, line 135: Possible to give an example of how the header looks like, I've taken a look at the spec but an actual example from something like a Moodle might be a better help in understanding.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
```diff
@@ -70,18 +71,27 @@ public async Task<IEnumerable<Member>> Get(string clientId, string tokenUrl, str
...
             _logger.LogInformation("Getting members.");
-            using var response = await httpClient.GetAsync(membershipUrl);
+            List<Member> members = new List<Member>();
 
-            if (!response.IsSuccessStatusCode)
+            do
             {
-                _logger.LogError("Could not get members.");
-                throw new Exception(response.ReasonPhrase);
+                using var response = await httpClient.GetAsync(membershipUrl);
```
adbindal, line 78: This seems to be making the calls one after the other which could lead to slowing down the overall response time of the API, how much latency increases this causes in the API? Can we check if there's a way to parameterize the NRPS request, so that we can increase the limit of students? The limit can then be changed as azure function parameters to allow different schools to set it according to their needs 🤔

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
```diff
@@ -111,5 +122,32 @@ public async Task<Member> GetById(string clientId, string tokenUrl, string audie
...
+            string[] urls = linkUrls.Split(',');
+            string nextUrl = urls.FirstOrDefault(u => u.IndexOf("\"next\"", StringComparison.OrdinalIgnoreCase) >=0 );
+
+            if (string.IsNullOrEmpty(nextUrl))
+            {
+                return null;
+            }
+
+            // Regex to get the url from next url excluding < and >; characters.
+            Regex reg = new Regex("(?<=<)(.*?)(?=>;)");
```
adbindal, line 143: Seems like we would be creating this regex for every membership call. Is it possible to move this to a class variable so that the complete Regex object is not created for every call and can be reused across them?

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
```diff
@@ -111,5 +122,32 @@ public async Task<Member> GetById(string clientId, string tokenUrl, string audie
...
+            Match matchResults = reg.Match(nextUrl);
```
adbindal, line 144: Looks like this method can throw. Should we guard it via a try catch? (from [Docs](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.match?view=net-6.0#system-text-regularexpressions-regex-match(system-string)))

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
adbindal: Just for own curiosity, the problem with the following regex is the fact that we might need to use groups to get the actual value, or something else?
```
Regex reg = new Regex("<(.*?)>; rel=next");
```

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
nehalilani: Maitree's response:
1. It will change client functionality
2. Twice in a launch (me call and get all users call)

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
nehalilani: example header:
"<https://DOMAIN/api/lti/courses/1261/names_and_roles?page=1&per_page=50>; rel=\"current\",<https://DOMAIN/api/lti/courses/1261/names_and_roles?page=2&per_page=50>; rel=\"next\",<https://DOMAIN/api/lti/courses/1261/names_and_roles?page=1&per_page=50>; rel=\"first\",<https://DOMAIN/api/lti/courses/1261/names_and_roles?page=2&per_page=50>; rel=\"last\""

nehalilani: Maitree's response: Wanted to perform case insensitive check, but string.contains does not support StringComparison.OrdinalIgnoreCase in the current .NET version. So used indexof for now.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
nehalilani: The spec says that the tool may send a parameter to specify the list size, but the platform may send more/less than the specified value, and that it is on the tool to handle this case with the next-link-header. So handling this case. But yes we can think of the above suggested way in case it becomes a bottleneck for users. The latency increase now will depend on the chunk size that the platform sends and the course size.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
nehalilani: yes, that regex would work as well but the current is also working just fine

nehalilani (approved):

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
adbindal: Ohh, makes sense. Thanks for explaining.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
adbindal: Sure. I was primarily worried about the latency of the API since it'll be calling downstream APIs one after the other. If the latency increase is not too much, I'm ok with the current design.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
adbindal: Thanks for validating this. The reason I placed this comment is that I've got confused with the look-ahead and that took me some time to figure out. Hence, was curious to understand the background on using lookahead in this particular case.

backend/Bindings/Edna.Bindings.LtiAdvantage/Services/NrpsClient.cs
adbindal: Makes sense. Thanks for explaining. Is it possible to get some idea about how much latency increase this would cause for a target class of let's say 100 students. I randomly picked up 100, the actual number can be picked up from our customers (@leestott - any thoughts on what would be a good number to benchmark against?)

